PR #19, opened 2020-03-02
Adding example for Invoke/OnInvoke

thebillkidy:
# Description

Adding an example on how to utilize the OnInvoke method from the proto bindings. This example helps users on how to create a gRPC server that can listen on direct method invocations.

## Issue reference

#20 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


examples/invoke/invoke-receiver.py
```diff
@@ -0,0 +1,40 @@
...
+# Start a gRPC client
+port = os.getenv('DAPR_GRPC_PORT')
+channel = grpc.insecure_channel(f"localhost:{port}")
+client = dapr_services.DaprStub(channel)
+print(f"Started gRPC client on DAPR_GRPC_PORT: {port}")
+
+# Our server methods
+class DaprClientServicer(daprclient_services.DaprClientServicer):
+    def OnInvoke(self, request, context):
+        # Return response to caller
```
youngbupark: I would add the code here to check if method is `my_method`. if not, I would return error e.g. "unsupported methods"

thebillkidy: Good suggestion! I'm working on adding this :) I would as well like to add a custom Packing / Unpacking part to it to make the example complete (rather than string encoding everywhere in the Any structure).

Would you by any accident know how to Unpack correctly? Currently I have this for Packing:

```python
def OnInvoke(self, request, context):
    a = Any()
    a.Pack(response_pb2.SampleResponse(action=123456))
    return a
```

And for Unpacking:

```python
response = client.InvokeService(dapr_messages.InvokeServiceEnvelope(id="method-receiver", method="my_method", data=data))
res = response_pb2.SampleResponse()
response.data.Unpack(res)
print(res)
```

But this is somehow not working... would love your thoughts on it :)

thebillkidy: @youngbupark done, added filter as well as Custom Data Packing / Unpacking

Issue #38, opened 2020-05-06
Serialize state value without base64 encoding


youngbupark:
## Describe the proposal

Serialize state value without base64 encoding

Issue #25, opened 2020-03-14
Add timer to actor


youngbupark:
## Describe the proposal

feature branch: https://github.com/dapr/python-sdk/tree/features/actor

Add timer to actor sdk.

Issue #24, opened 2020-03-14
Add reminder to actor sdk


youngbupark:
## Describe the proposal

feature branch: https://github.com/dapr/python-sdk/tree/features/actor

Add reminder to actor sdk.

PR #39, opened 2020-05-10
Add timer and reminder and refactor codes

youngbupark:
# Description

**This PR will be the final PR to implement all actor features which dotnet actor and java actor sdks have!**

* Add Actor Timer and Actor Reminder
* Add Timer and Reminder handlers to flask_dapr extension
* Add Timer and reminder example to demo_actor app
* Use JSON serializer for state value serialization instead of using base64 encoding - adding json serialization for bytes type; unlike dotnet/java json serializer, default python json serializer doesn't support byte array. 
* Make coding style consistent
* Handle Exceptions via DaprInternalError Exception (which is similar to DaprException in dotnet-sdk)
* Add lots of unit-tests for reminder/timer codes

## Issue reference

#25 #24 #38 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


dapr/actor/runtime/stateprovider.py
```diff
@@ -78,14 +76,14 @@ async def save_state(
             json_output.write(operation)
             json_output.write(b'","request":{"key":"')
             json_output.write(state.state_name.encode('utf-8'))
+            json_output.write(b'"')
             if state.value is not None:
                 serialized = self._state_serializer.serialize(state.value)
```
youngbupark, line 81: @artursouza As you pointed it out before, I changed the code to not encode the data in base64. Please double-check if it is expected or not.

cc/ @amanbha

dapr/actor/runtime/reminderdata.py
```diff
@@ -0,0 +1,64 @@
...
+        return self._period
+
+    def as_dict(self) -> dict:
+        encoded_state = None
+        if self._state is not None:
+            encoded_state = base64.b64encode(self._state)
+        return {
+            'name': self._name,
+            'dueTime': self._due_time,
+            'period': self._due_time,
```
amanbha, line 54: does this get serialized into the period format as required by dapr runtime  1h5m5s?

youngbupark: yes json serializer converts to golang format.  please see the code and tests:

https://github.com/dapr/python-sdk/blob/6f07dd55a529c7782bc9703ce28c3da63fa83f3d/dapr/serializers/json.py#L66-L67

https://github.com/dapr/python-sdk/blob/youngp/add-timer/tests/serializers/testutil.py

Issue #73, opened 2020-06-22
Add Invoke service api to convenient layer


youngbupark:
## Describe the proposal

Add Invoke service api to convenient layer.

Parent epic : #67 

- [x] Implementation
- [x] Unit-tests
- [x] Documentation in code
- [x] update examples

PR #78, opened 2020-06-25
Initial commit of invoke_service

youngbupark:
# Description

This is the initial commit of invoke_service. With new convenient layer, user needs only three lines of codes without the knowledge of gRPC.

```python
>>> import dapr
>>> with dapr.Dapr() as d:
...     resp = d.invoke_service('callee', 'method', b'data')
...     resp.bytesdata
b'return_data'
>>>
```

Change summary:

- Defined `Dapr` class for dapr client.
- Added invoke_service.
- Added unit-tests.
- Fixed import cycle in actor.
- Updated invoke-simple and invoke-custom with new api.

## Issue reference

#73 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


youngbupark:
@chinzhiweiblank JFYI - Even if this is WIP PR, please give any feedbacks.

dapr/clients/grpc/dapr_client.py
```diff
@@ -0,0 +1,168 @@
...
+    @property
+    def proto(self) -> GrpcAny:
+        return self._data
+
+    @property
+    def content_type(self) -> Optional[str]:
+        return self._content_type
+
+
+class DaprResponse:
```
youngbupark, line 50: @shalabhms I have thought about how we can retrieve trace id easily. my initial idea is to have `DaprResponse` object which allows users to get any headers. trace + span id will be one of headers.. so that they can use it.

But. we need to discuss it more.

shalabhms: @youngbupark , yes thats the right direction. I was having similar discussion with @amanbha , as how to do it in dot net sdk. we can discuss it more on this also we have to think a way/expose a helper that user can pass any metadata.

Issue #91, opened 2020-07-10
Support DAPR_API_TOKEN


youngbupark:
## Describe the proposal

This is to support dapr api token authentication.  - Reference: https://github.com/dapr/docs/blob/master/howto/enable-dapr-api-token-based-authentication/README.md

1. read DAPR_API_TOKEN environment variable
2. set `dapr-api-token` metadata in each api call.

Use grpc client interceptor.

PR #95, opened 2020-07-13
Add DaprClientInterceptor to add dapr-api-token header

mukundansundar:
# Description

Add a client interceptor usin the UnaryUnaryClientInterceptor.

Referring to the API definition here: https://grpc.github.io/grpc/python/grpc.html#client-side-interceptor

This API is mentioned as an **Experimental API for python gRPC sdk**

## Issue reference

Please reference the issue this PR will close: #91 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


dapr/clients/grpc/_helpers.py
```diff
@@ -6,6 +6,98 @@
...
+            ['method', 'timeout', 'metadata', 'credentials', 'wait_for_ready', 'compression']),
+        ClientCallDetails):
+    """This is an implementation of the ClientCallDetails interface needed for interceptors.
+    This class takes six named values and inherits the ClientCallDetails from grpc package.
+    This class encloses the values that describe a RPC to be invoked.
+    """
+    pass
+
+
+class HeaderInterceptor(UnaryUnaryClientInterceptor):
```
youngbupark, line 29: Later, this can be used by trace id injection, so that I would rename `HeaderInterceptor` to `DaprClientInterceptor`.

tests/clients/test_dapr_grpc_client.py
```diff
@@ -150,6 +152,27 @@ def test_publish_error(self):
...
+            content_type="text/plain",
+            metadata=(
+                ('key1', 'value1'),
+                ('key2', 'value2'),
+            ),
+        )
+
+        self.assertEqual(b'haha', resp.content)
+        self.assertEqual("text/plain", resp.content_type)
+        self.assertEqual(4, len(resp.headers))
```
pruthvidhodda, line 171: This should be 3 right? The metadata from request (2) and 1 dapr-api-token.

mukundansundar: No it is 4. I have seen that it is always 1 more than we expect. `user-agent` is added automatically

Issue #109, opened 2020-08-17
Add state transaction api


youngbupark:
## Describe the proposal

Add state transaction api - https://github.com/dapr/dapr/pull/1892

PR #116, opened 2020-09-18
Add transaction state API.

artursouza:
# Description

Add transaction state API.

## Issue reference

Please reference the issue this PR will close: #109 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/dapr/clients/grpc/_request.py
+++ b/dapr/clients/grpc/_request.py
@@ -5,6 +5,7 @@ Copyright (c) Microsoft Corporation.
 Licensed under the MIT License.
 """
 
+from enum import Enum
 from typing import Dict, Optional, Union
 
 from google.protobuf.any_pb2 import Any as GrpcAny
@@ -256,3 +257,65 @@ class BindingRequest(DaprRequest):
     def binding_metadata(self):
         """Gets the metadata for output binding."""
         return self._binding_metadata
+
+
+class TransactionOperationType(Enum):
+    """Represents the type of operation for a Dapr Transaction State Api Call"""
+    upsert = "upsert"
+    delete = "delete"
+
+
+class TransactionalStateOperation:
+    """An upsert or delete operation for a state transaction, 'upsert' by default.
+
+    Attributes:
+        key (str): state's key.
+        data (Union[bytes, str]): state's data.
+        etag (str): state's etag.
+        operation_type (TransactionOperationType): operation to be performed.
+    """
+
+    def __init__(
+            self,
+            key: str,
+            data: Union[bytes, str],
+            etag: str = '',
+            operation_type: TransactionOperationType = TransactionOperationType.upsert):
+        """Initializes TransactionalStateOperation item from :obj:`runtime_v1.TransactionalStateOperation`.
+
+        Args:
+            key (str): state's key.
+            data (Union[bytes, str]): state's data.
+            etag (str): state's etag.
+            operationType (Optional[TransactionOperationType]): operation to be performed.
+
+        Raises:
+            ValueError: data is not bytes or str.
+        """
+        if not isinstance(data, (bytes, str)):
+            raise ValueError(f'invalid type for data {type(data)}')
+
+        self._key = key
+        self._data = data  # type: ignore
+        self._etag = etag
+        self._operation_type = operation_type
+
+    @property
+    def key(self) -> str:
+        """Gets key."""
+        return self._key
+
+    @property
+    def data(self) -> Union[bytes, str]:
+        """Gets raw data."""
+        return self._data
+
+    @property
+    def etag(self) -> str:
+        """Gets etag."""
+        return self._etag
+
+    @property
+    def operation_type(self) -> TransactionOperationType:
+        """Gets etag."""
+        return self._operation_type
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -479,9 +479,7 @@ class DaprClient:
                             operation_type=TransactionOperationType.delete,
                             key=key_to_delete),
                     ],
-                    transactional_metadata=(
-                        ('header1', 'value1')
-                    ),
+                    transactional_metadata={"header1": "value1"},
                     metadata=(
                         ('header1', 'value1')
                     ),
@@ -490,7 +488,7 @@ class DaprClient:
         Args:
             store_name (str): the state store name to save to.
             operations (Sequence[TransactionalStateOperation]): the transaction operations.
-            transactional_metadata (tuple, optional): custom metadata for transaction request.
+            transactional_metadata (Dict[str, str], optional): custom metadata for transaction.
             metadata (tuple, optional): custom grpc metadata.
 
         Returns:
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -16,7 +16,11 @@ from dapr.conf import settings
 from dapr.proto import api_v1, api_service_v1, common_v1
 
 from dapr.clients.grpc._helpers import MetadataTuple, DaprClientInterceptor, to_bytes
-from dapr.clients.grpc._request import InvokeServiceRequest, BindingRequest
+from dapr.clients.grpc._request import (
+    InvokeServiceRequest,
+    BindingRequest,
+    TransactionalStateOperation,
+)
 from dapr.clients.grpc._response import (
     BindingResponse,
     DaprResponse,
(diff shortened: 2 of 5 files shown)


dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,59 @@ def save_state(
...
+        """Saves key-value pairs to a statestore as a transaction
+
+        This saves key-values to the statestore as part of a single transaction
+        Options for request can be passed with the options field and custom
+        metadata can be passed with metadata field.
+
+        The example saves states to a statestore:
+            from dapr import DaprClient
+            with DaprClient() as d:
+                resp = d.save_state(
```
mukundansundar, line 472: incorrect doc string.

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,59 @@ def save_state(
...
+            request=common_v1.StateItem(
+                key=o.key,
+                value=to_bytes(o.data),
+                etag=o.etag,
+                metadata=o.metadata)) for o in operations]
+
+        req = api_v1.ExecuteStateTransactionRequest(
+            storeName=store_name,
+            operations=req_ops,
+            metadata=metadata)
```
mukundansundar, line 506: is the same metadata passed in both the gRPC call and here? @youngbupark ?

tests/clients/test_dapr_grpc_client.py
```diff
@@ -198,29 +199,20 @@ def test_get_save_delete_state(self):
         resp = dapr.get_state(store_name="statestore", key=key)
         self.assertEqual(resp.data, b'')
 
-    def test_save_get_states(self):
+    def test_transaction_then_get_states(self):
```
mukundansundar, line 202: am I missing something or why is save_state test removed?

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,59 @@ def save_state(
         return DaprResponse(
             headers=call.initial_metadata())
 
+    def execute_transaction(
+            self,
+            store_name: str,
+            operations: Sequence[TransactionalStateOperation],
+            metadata: Optional[MetadataTuple] = ()) -> DaprResponse:
```
mukundansundar, line 462: I would prefer to have this a mandatory field. 
@youngbupark thoughts?

mukundansundar (changes requested):

tests/clients/test_dapr_grpc_client.py
artursouza: It is in another test. I decided to bundle it in the bulk get test since it seems they fit well together.

dapr/clients/grpc/client.py
artursouza: Forgot that, adding it now.

dapr/clients/grpc/client.py
youngbupark: python-sdk exposes metadata only for grpc request. it is definitely optional so that user can pass the custom metadata to the grpc request. 

but we need to add another metadata for `transactional_metadata` for metadata in transactional request. it is also optional.

dapr/clients/grpc/client.py
artursouza: Good point, what do we do with the metadata from proto and the one in the call? Reuse or keep them separate.

dapr/clients/grpc/client.py
artursouza: OK. Adding transactional_metadata.

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,56 @@ def save_state(
...
+        """
+        if not store_name or len(store_name) == 0 or len(store_name.strip()) == 0:
+            raise ValueError("State store name cannot be empty")
+        req_ops = [api_v1.TransactionalStateOperation(
+            operationType=o.operation_type.value,
+            request=common_v1.StateItem(
+                key=o.key,
+                value=to_bytes(o.data),
+                etag=o.etag,
+                metadata=o.metadata)) for o in operations]
```
youngbupark, line 498: we do not support per operation metadata for transactional request.

dapr/clients/grpc/client.py
youngbupark: We need set metadata in `TransactionalStateOperation` message. As mentioned above, this metadata is `transactional_metadata` which is List type, not Tuple.

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,56 @@ def save_state(
...
+        This saves or deletes key-values to the statestore as part of a single transaction,
+        metadata can be passed with metadata field.
+
+        The example saves states to a statestore:
+            from dapr import DaprClient
+            with DaprClient() as d:
+                resp = d.execute_transaction(
+                    store_name='state_store',
+                    operations=[
+                        TransactionalStateOperation(key=key, data=value),
```
youngbupark, line 474: I think operation type is missing in this example.

examples/state_store/state_store.py
```diff
@@ -24,6 +27,12 @@
     data = d.get_state(store_name=storeName, key=key).data
     print(f"Got value: {data}")
 
+    # Transaction upsert
+    d.execute_transaction(store_name=storeName, operations=[
+        TransactionalStateOperation(key=key, data=updated_value),
```
youngbupark, line 32: operation type needs to be set for each operation

youngbupark (changes requested):

dapr/clients/grpc/client.py
artursouza: Using separate metadata argument for grpc vs request.

dapr/clients/grpc/client.py
artursouza: We do, it is in the proto.

dapr/clients/grpc/client.py
artursouza: It defaults to 'upsert'.

examples/state_store/state_store.py
artursouza: It defaults to 'upsert', do you think it should be required? I found it obvious that upsert is the most common scenario.

youngbupark: default upsert is ok but I think we need to show the usage of the API with using possible options in the test and docstring by default . Default operation type behavior scenario would be additional example.

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,64 @@ def save_state(
...
+                        TransactionalStateOperation(key=another_key, data=another_value),
+                        TransactionalStateOperation(
+                            operation_type=TransactionOperationType.delete,
+                            key=key_to_delete),
+                    ],
+                    transactional_metadata=(
+                        ('header1', 'value1')
+                    ),
+                    transactional_metadata=(
```
mukundansundar, line 485: additional field

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,64 @@ def save_state(
...
+                        ('header1', 'value1')
+                    ),
+                )
+
+        Args:
+            store_name (str): the state store name to save to.
+            operations (Sequence[TransactionalStateOperation]): the transaction operations.
+            metadata (tuple, optional): custom metadata.
```
mukundansundar, line 493: transcational_metadata

mukundansundar (dismissed):

dapr/clients/grpc/_request.py
```diff
@@ -256,3 +257,74 @@ def text(self) -> str:
...
+    """
+
+    def __init__(
+            self,
+            key: str,
+            data: Union[bytes, str],
+            etag: str = '',
+            operation_type: TransactionOperationType = TransactionOperationType.upsert,
+            metadata: Optional[MetadataTuple] = ()):
+        """Initializes BulkStateItem item from :obj:`runtime_v1.BulkStateItem`.
```
youngbupark, line 286: docstring - this isn't updated :) this must be TransactionalStateOperation :)

youngbupark (changes requested):

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,65 @@ def save_state(
         return DaprResponse(
             headers=call.initial_metadata())
 
+    def execute_transaction(
+            self,
+            store_name: str,
+            operations: Sequence[TransactionalStateOperation],
+            transactional_metadata: Optional[MetadataTuple] = (),
```
youngbupark, line 462: This is not tuple. it is `Dict[str, str]` type
```
transactional_metadata: Dict[str, str] = {}
```

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,64 @@ def save_state(
...
+                    store_name='state_store',
+                    operations=[
+                        TransactionalStateOperation(key=key, data=value),
+                        TransactionalStateOperation(key=another_key, data=another_value),
+                        TransactionalStateOperation(
+                            operation_type=TransactionOperationType.delete,
+                            key=key_to_delete),
+                    ],
+                    transactional_metadata=(
+                        ('header1', 'value1')
```
mukundansundar, line 483: change to dict value

dapr/clients/grpc/client.py
```diff
@@ -451,6 +455,64 @@ def save_state(
...
+                    ),
+                    metadata=(
+                        ('header1', 'value1')
+                    ),
+                )
+
+        Args:
+            store_name (str): the state store name to save to.
+            operations (Sequence[TransactionalStateOperation]): the transaction operations.
+            transactional_metadata (tuple, optional): custom metadata for transaction request.
```
mukundansundar, line 493: dict[str,str]

mukundansundar (changes requested):

mukundansundar (approved):

youngbupark (approved):

PR #118, opened 2020-09-21
Adding metadata to applicable state APIs.

artursouza:
# Description

Adding metadata to applicable state APIs.

## Issue reference

Please reference the issue this PR will close: #114 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -313,6 +313,7 @@ class DaprClient:
             self,
             store_name: str,
             key: str,
+            state_metadata: Optional[Dict[str, str]] = dict(),
             metadata: Optional[MetadataTuple] = ()) -> DaprResponse:
         """Gets value from a statestore with a key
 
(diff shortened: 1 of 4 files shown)


mukundansundar (commented):
add unit tests?

youngbupark (approved):

youngbupark:
> add unit tests?

oh that's good point :)

artursouza:
Added unit tests.

mukundansundar (commented):
I am able to see that save_state has metadata in the StateItem [definition](https://github.com/dapr/dapr/blob/master/dapr/proto/common/v1/common.proto#L95).
Do we not use it? 
It is documented [here](https://github.com/dapr/docs/blob/master/reference/api/state_api.md#save-state) that metadata can be passed, but we are not passing along the metadata for state item in the save_state call.

artursouza:
> here

Thanks for pointing that out. I did not find the metadata in the request object, so I thought it was not applicable.

mukundansundar (approved):
Do we need to provide a convenience method to save multiple state? @youngbupark

artursouza:
> Do we need to provide a convenience method to save multiple state? @youngbupark

This is another scope. I noticed this too. I will merge this PR in the meantime.

youngbupark:
> > Do we need to provide a convenience method to save multiple state? @youngbupark
> 
> This is another scope. I noticed this too. I will merge this PR in the meantime.

That's the another issue. but the other sdk supports multiple states save ?

PR #132, opened 2020-10-15
update pkg install instruction in READMEs

### LINKED ISSUE #125: Pip install dapr does not install all the correct packages  in non-development mode
## Expected Behavior

pip install dapr should install all the correct modules. Currently it looks like there are some modules that you only get when you do pip install dapr-dev that are being used in samples. Why are there different packages herehttps://github.com/dapr/python-sdk#install-dapr-python-sdk which make it confusing?

## Actual Behavior

Samples fail with **dapr.ext.grpc module not found**

## Steps to Reproduce the Problem

On a new machine do 
**pip install dapr**

Then run pub/sub sample https://github.com/dapr/python-sdk/tree/master/examples/pubsub-simple

Fails with **dapr.ext.grpc module not found**

Also what is the dapr.proto.grpc module do?

youngbupark:
# Description

update pkg install docs

## Issue reference

closes #125 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


README.md
```diff
@@ -17,7 +17,9 @@ This includes the following packages:
 * [dapr.conf](./dapr/conf): Configuration
 * [dapr.serializers](./dapr/serializers): serializer/deserializer
 * [dapr.proto](./dapr/proto): Dapr gRPC autogenerated gRPC clients
-* [flask_dapr](./flask_dapr): flask extension for Dapr
+* [dapr.ext.grpc](./ext/dapr-ext-grpc): gRPC extension for Dapr
+* [dapr.ext.fastapi](./ext/dapr-ext-fastapi): FastAPI extension (actor) for Dapr
+* [flask.dapr](./ext/flask-dapr): Flask extension (actor) for Dapr
```
mukundansundar, line 22: is there any reason why flask_dapr is different from the other 2 ext modules? Can they be made uniform?

youngbupark: We need to deprecate this one. Flask doesn't support asynchronous method calls correctly but actor is written asynchronously. I saw some race condition issues in hackathon project when it runs in multi-thread scenarios.

User needs to use fastapi extension to use python-actor. I will delete it after this pr.

mukundansundar (approved):

Issue #150, opened 2020-12-10
Naming changes


yaron2:
To keep consistent with other SDK changes per maintainers overview, the following changes need to be made:

**Service Invocation:**

* Old method: **invoke_service**
* New method: **invoke_method**

<br>
<br>

* Old parameter: **id**
* New parameter: **app_id**

* Old parameter: **method**
* New parameter: **method_name**

**State**

* Old method: **saves_states**
* New method: **save_bulk_state**

* Old method: **get_states**
* New method: **get_bulk_state**

* Old method: **execute_transaction**
* New method: **execute_state_transaction**

* Old method: **saves_states**
* New method: **save_bulk_state**

**Output Bindings**

* Old parameter: **name**
* New parameter: **binding_name**

**Pub Sub**

* Old parameter: **topic**
* New parameter: **topic_name**

** Actors Proxy**

* Old method: **invoke**
* New method: **invoke_method**

** Actor Reminders**

* Old parameter: **name**
* New parameter: **reminder_name**

** Actor Timers**

* Old parameter: **name**
* New parameter: **timer_name**

PR #168, opened 2021-01-14
Naming changes for consistency across SDKs

karishma-chawla:
# Description

Updated API and parameter names for consistency across SDKs.

## Issue reference

https://github.com/dapr/python-sdk/issues/150
https://github.com/dapr/dapr/issues/2535

Please reference the issue this PR will close: #150

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


karishma-chawla:
Added changes to tox.ini file by mistake, reverting them.

dapr/actor/runtime/_reminder_data.py
```diff
@@ -72,7 +72,7 @@ def as_dict(self) -> Dict[str, Any]:
         if self._state is not None:
             encoded_state = base64.b64encode(self._state)
         return {
-            'name': self._name,
+            'reminderName': self._reminder_name,
```
mukundansundar, line 75: @yaron2 @youngbupark Similar to this should `timerName` be returned in `as_dict` call for the timer class seen below?
See line 79 in `_timer_data.py` file.

dapr/clients/grpc/_response.py
```diff
@@ -66,7 +66,7 @@ def get_headers(self, as_dict: bool = False) -> Union[MetadataDict, MetadataTupl
 
 
 class InvokeServiceResponse(DaprResponse):
```
mukundansundar, line 68: @yaron2 Should this be renamed to InvokeMethodResponse? and similar changes for other classes?

dapr/proto/common/v1/common_pb2_grpc.py
```diff
@@ -1,3 +1,4 @@
 # Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+"""Client and server classes corresponding to protobuf-defined services."""
```
mukundansundar, line 2: @karishma-chawla was this automatically added?

dapr/clients/grpc/_response.py
karishma-chawla: I actually did rename to InvokeMethodResponse/Request, only to realize while testing that we have defined them as InvokeService* in Dapr runtime https://github.com/dapr/java-sdk/blob/6a44d51279d7d6d5ccdfd63346b4c83d00e7eeb8/sdk/src/main/java/io/dapr/client/DaprClient.java, so ended up changing it back.

Jave SDK has also similar naming convention right now https://github.com/dapr/java-sdk/blob/9dcff43ef08497b7169b93207002f450a76236bd/sdk/src/main/java/io/dapr/client/domain/InvokeServiceRequest.java

I do think though that we should rename everything including runtime to InvokeMethod* for consistency, but not sure if it should be prioritized over other things.

dapr/proto/common/v1/common_pb2_grpc.py
karishma-chawla: I ran the script in the repo for generated code to let it pick latest changes https://github.com/dapr/python-sdk/blob/master/tools/regen_grpcclient.sh

dapr/clients/grpc/_response.py
karishma-chawla: I noticed that Artur updated Java SDK now to use InvokeMethodResponse/Request, so I've updated the PR to reflect that in python SDK as well. Calls to runtime are still InvokeServiceResponse as I mentioned in the previous comment, just a heads up.

artursouza (approved):

PR #175, opened 2021-01-25
Bootstrapped autovalidation for python examples

wcs1only:
# Description

Add autovalidation to examples/pubsub-simple and examples/state_store

## Issue reference

ref: #177

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/workflows/validate_examples.yaml
+++ b/.github/workflows/validate_examples.yaml
@@ -1,0 +1,85 @@
+name: validate-examples
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - master
+      - release-*
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    env:
+      PYTHON_VER: 3.7
+      GOVER: 1.15.3
+      GOOS: linux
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yaml
```diff
@@ -18,47 +18,105 @@ jobs:
...
+      - name: Check Typing
+        run: |
+          tox -e type
+      - name: Run unit-tests
+        run: |
+          tox -e py37
+      - name: Check Examples
+        run: |
+          tox -e examples
+      - name: Upload test coverage
```
mukundansundar, line 98: can be done before examples.

mukundansundar (commented):
quick question should the validation be done in a separate workflow similar to quickstarts?

.github/workflows/build.yaml
wcs1only: > quick question should the validation be done in a separate workflow similar to quickstarts?

It could. I mostly did it this way for expediency, but it would probably be cleaner as a separate workflow. Let me split it out.

mukundansundar (approved):

artursouza (approved):

Issue #176, opened 2021-01-26
Makes service method API invocations over HTTP by default.


wcs1only:
Today, mapping gRPC to HTTP is not correct and we don't have a mapping that can work yet (as per dapr/dapr#2342). So, all API calls should be made over HTTP by default in the SDK, even if the other APIs work over gRPC.

This should improve experience for most users since HTTP endpoint is most common. The gRPC to gRPC or HTTP to gRPC scenarios are way less common since dapr requires a custom proto to be implemented in the app.

Parent: dapr/dapr#2342
See also: dapr/java-sdk#450

## Release Note

RELEASE NOTE:

PR #182, opened 2021-01-28
Add an HTTP client for service invocation and default to using it

wcs1only:
# Description

Created an HTTP client to use for service invocation. Also added a configuration option allow GRPC to be used as well.

> **Note:** This PR was getting quite long, so I've split out another issue to add tracing.

## Issue reference

Please reference the issue this PR will close: #176

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


dapr/clients/__init__.py
```diff
@@ -7,7 +7,7 @@
 
 from dapr.clients.base import DaprActorClientBase
 from dapr.clients.exceptions import DaprInternalError, ERROR_CODE_UNKNOWN
-from dapr.clients.grpc.client import DaprClient
+from dapr.clients.generic import DaprClient
```
artursouza, line 10: `generic` is an overloaded term here.  Can you name this package something else? Maybe `main`?

dapr/clients/generic.py
```diff
@@ -0,0 +1,64 @@
...
+            app_id: str,
+            method_name: str,
+            data: Union[bytes, str, GrpcMessage],
+            content_type: Optional[str] = None,
+            metadata: Optional[MetadataTuple] = None,
+            http_verb: Optional[str] = None,
+            http_querystring: Optional[MetadataTuple] = None) -> InvokeMethodResponse:
+
+        if self.invocation_client:
+            return self.invocation_client.invoke_method(
```
artursouza, line 48: Why tests do not cover this if-else?

dapr/clients/grpc/_response.py
```diff
@@ -191,6 +191,11 @@ def unpack(self, message: GrpcMessage) -> None:
             ValueError: message is not protocol buffer message object or message's type is not
                 matched with the response data type
         """
+
+        if self.content_type == "application/x-protobuf":
```
artursouza, line 195: handle lower case vs upper case and content type optional suffix (`application/x-protobuf; gzip`)

dapr/clients/grpc/client.py
```diff
@@ -38,7 +38,7 @@
 from opencensus.trace.tracers.base import Tracer   # type: ignore
 
 
-class DaprClient:
+class DaprGrpcClient:
```
artursouza, line 41: Why renaming it if it is in another package?

dapr/clients/http/client.py
```diff
@@ -0,0 +1,71 @@
...
+            error_info = self._serializer.deserialize(error_body)
+        except Exception:
+            return DaprInternalError(f'Unknown Dapr Error. HTTP status code: {response.status}')
+
+        if error_info and isinstance(error_info, dict):
+            message = error_info.get('message')
+            error_code = error_info.get('errorCode') or ERROR_CODE_UNKNOWN
+            return DaprInternalError(message, error_code)
+
+        return DaprInternalError(f'Unknown Dapr Error. HTTP status code: {response.status}')
```
artursouza, line 71: Why DaprInternalError and not just DaprError? Internal means customers will not get this, is this correct?

dapr/serializers/json.py
```diff
@@ -28,6 +27,9 @@ def serialize(
             custom_hook: Optional[Callable[[object], bytes]] = None) -> bytes:
 
         dict_obj = obj
+
+        # importing this from top scope creates a circular import
+        from dapr.actor.runtime.config import ActorRuntimeConfig
```
artursouza, line 32: Wait. Why we did not have this before?

tests/clients/test_dapr_grpc_client.py
```diff
@@ -11,7 +11,7 @@
 
 from unittest.mock import patch
 
-from dapr.clients.grpc.client import DaprClient
+from dapr.clients.grpc.client import DaprGrpcClient as DaprClient
```
artursouza, line 14: This is odd because it was named correctly before.

artursouza (changes requested):

dapr/clients/generic.py
wcs1only: I think the gRPC unit tests include the gRPC client class directly.

dapr/serializers/json.py
wcs1only: We did, I just moved it out of the module scope.

dapr/clients/grpc/client.py
wcs1only: Because there is a new DaprClient which inherits from this class. Would be confusing if I named it the same, wouldn't it?

dapr/clients/http/client.py
wcs1only: This is an existing exception. If I changed the name, that would be a breaking change.

dapr/clients/__init__.py
wcs1only: Yeah, can do.

dapr/clients/grpc/_response.py
wcs1only: The http library already correctly parses out the suffix for us, but I'll add a test to confirm this. Will also make case sensitive.

dapr/clients/grpc/client.py
artursouza: You can name it differently when importing it. It is already in another package.

dapr/serializers/json.py
artursouza: You mean we had a circular import before?

tests/clients/test_dapr_grpc_client.py
artursouza: This is what I mean, you can just name it DaprGrpcClient when using it. So the class can keep the original name.

dapr/clients/http/dapr_invocation_http_client.py
```diff
@@ -0,0 +1,75 @@
...
+            http_querystring: Optional[MetadataTuple] = None) -> InvokeMethodResponse:
+        """This implementation is designed to seemlessly mimic the behavior of the grpc
+        service invoke implementation. Please see the grpc version for parameter definitions
+        and samples"""
+
+        verb = 'GET'
+        if http_verb is not None:
+            verb = http_verb
+
+        headers = {}
```
artursouza, line 41: We need DAPR_API_TOKEN and the traceparent header to be set here.

dapr/clients/grpc/client.py
wcs1only: Yeah, but we only named it so generically in the first place because we only had the gRPC client and nothing else. Now that we have both gRPC and HTTP, naming them more specifically makes sence.

wcs1only:
Tracing and API token added. I've verified the sample. Everything looks right, though the grpc integration has an extra span in the stack.

tests/clients/test_dapr_grpc_client.py
wcs1only: Yeah, but I'd rather not have two classes which are both called DaprClient in the overall project, even if I can alias one of them away. Prefer to just call the DaprGrpcClient the same everywhere and use the generic name DaprClient only for the public facing class that does both Grpc and HTTP. I've updated this to not use an alias, rather just use the name DaprGrpcClient everywhere, so there is no ambiguity anywhere.

artursouza (approved):

Issue #183, opened 2021-01-30
Update python-sdk to allow setting of data_content_type for cloud event envelope.


wcs1only:
Python-sdk must support dapr/components-contrib#571

## Release Note

RELEASE NOTE: **ADD** Add data_content_type parameter to publish_event

PR #185, opened 2021-02-01
Add python-sdk support for publisher cloudevent content_type

wcs1only:
# Description

Python-sdk did not have a way to set data_content_type on a publish event. 

## Issue reference

Please reference the issue this PR will close: #183

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -289,7 +289,8 @@ class DaprGrpcClient:
             pubsub_name: str,
             topic_name: str,
             data: Union[bytes, str],
-            metadata: Optional[MetadataTuple] = ()) -> DaprResponse:
+            metadata: Optional[MetadataTuple] = (),
+            data_content_type: Optional[str] = None) -> DaprResponse:
         """Publish to a given topic.
         This publishes an event with bytes array or str data to a specified topic and
         specified pubsub component. The str data is encoded into bytes with default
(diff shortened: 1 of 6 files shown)


dapr/clients/grpc/client.py
```diff
@@ -315,6 +316,7 @@ def publish_event(
             topic_name (str): the topic name to publish to
             data (bytes or str): bytes or str for data
             metadata (tuple, optional): custom metadata
+            data_content_type: (str, optional): content type of the data payload
```
mukundansundar, line 319: should this made default since it is not set by the SDK?

wcs1only: CloudEvent spec treats this as optional:
https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#optional-attributes

As does a comment in our own proto spec.

dapr/clients/grpc/client.py
```diff
@@ -289,7 +289,8 @@ def publish_event(
             pubsub_name: str,
             topic_name: str,
             data: Union[bytes, str],
-            metadata: Optional[MetadataTuple] = ()) -> DaprResponse:
+            metadata: Optional[MetadataTuple] = (),
+            data_content_type: Optional[str] = None) -> DaprResponse:
```
artursouza, line 293: Default to application/json instead of None.

wcs1only: Why? We can take a blob here, and this is an optional field in the cloud event spec?

artursouza: OK. Let the runtime decide on the default behavior.

artursouza (approved):

mukundansundar (approved):

PR #206, opened 2021-03-21
feat: add shutdown

Hugome:
# Description

Add Shutdown action

## Issue reference

Related to dapr/dapr#2828

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


wcs1only:
Unit tests are failing, probably just need a shutdown implementation in tests/clients/fake_dapr_server.py ?

wcs1only:
> Unit tests are failing, probably just need a shutdown implementation in tests/clients/fake_dapr_server.py ?

I sent you a PR @Hugome with a patch that makes the unit tests pass:
https://github.com/Hugome/python-sdk/pull/1

Hugome:
Merged, thanks for the fix :)

dapr/clients/grpc/client.py
```diff
@@ -809,3 +810,22 @@ def wait(self, timeout_s: float):
...
+
+        This will ask the sidecar to gracefully shutdown.
+
+        The example shutdown the sidecar:
+
+            from dapr.clients import DaprClient
+
+            with DaprClient() as d:
+                resp = d.shutdown()
+        """
```
wcs1only, line 827:
```suggestion

        Args:
            metadata (tuple, optional): custom grpc metadata
        """
```

Though I'm not sure it is strictly necessary to take metadata as a parameter at all here. We don't do anything with it on the runtime side, correct?

wcs1only:
Marking this do-not-merge pending dapr/dapr#2828. If that makes it into the v1.1.0 release, we can safely merge this PR. Otherwise, we'll wait until after v1.1.0 release.

dapr/clients/grpc/client.py
Hugome: I was not sure for the metadata, does the auth token (https://docs.dapr.io/operations/security/api-token/) can go through those metadata ? Or it's managed by the `DaprClient` ?

wcs1only: We have an interceptor add the token to every request: 
https://github.com/dapr/python-sdk/blob/master/dapr/clients/grpc/client.py#L89-L93

The user shouldn't need to pass anything themselves.

wcs1only (approved):

Issue #162, opened 2021-01-06
Set the timeout for ActorProxy


alydersen:
Hi everybody,

We have a few issues regarding errors and timeouts when talking to the actors that we are ironing out. The question is if it its possible to tweak the default timeout of 60 seconds on the ActorProxy or otherwise instantialize the class? I guessing there will be very few cases where 60 seconds is needed?

Thanks!
Andreas

RELEASE NOTE: **ADD** Add DAPR_HTTP_TIMEOUT_SECONDS configuration option.

wcs1only:
We can indeed add a timeout value to the public interface.

wcs1only:
So we're adding two ways to do this. You can set the environment variable DAPR_HTTP_TIMEOUT_SECONDS

or

```python
proxy = ActorProxyFactory(http_timeout_seconds=10).create('actor_type', 'actor_id')
```

PR #207, opened 2021-03-23
Add a configuration to specify http timeouts (#162)

wcs1only:
# Description

Add a configuration option DAPR_HTTP_TIMEOUT_SECONDS. This can be set as an environment variable, or it can be passed as a parameter to ActorProxyFactory(), or DaprClient().

## Issue reference

Please reference the issue this PR will close: #162

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


dapr/clients/__init__.py
```diff
@@ -55,14 +56,18 @@ def __init__(
...
         """
         super().__init__(address, interceptors)
         self.invocation_client = None
 
         invocation_protocol = settings.DAPR_API_METHOD_INVOCATION_PROTOCOL.upper()
 
         if invocation_protocol == 'HTTP':
-            self.invocation_client = DaprInvocationHttpClient(headers_callback=headers_callback)
+            if http_timeout_seconds is None:
+                http_timeout_seconds = settings.DAPR_HTTP_TIMEOUT_SECONDS
```
artursouza, line 68: Can the default be handles in the param in the method signature?

wcs1only: Well, I initially had it that way, but in order to be able to manipulate settings for the sake of unit testing, I had to use this approach. 

TIL: python method signature defaults are set once with the class is loaded and are immutable after that.

artursouza (approved):

Issue #214, opened 2021-04-20
Add support for actor reentrancy


halspang:
## Describe the feature
The dapr core runtime is planning on supporting actor reentrancy. The bulk of this logic will be handled in dapr's runtime but the SDK will be responsible for:

- Preserving the header `dapr-reentrant-id`
- Providing the new config fields required by `dapr/config`

For more context on actor reentrancy, the design proposal can be found here: [actor reentrancy](https://github.com/dapr/dapr/issues/2901)

## Release Note
RELEASE NOTE: **ADD** Support for actor reentrancy

PR #219, opened 2021-05-14
Adds support for Actor Reentrancy

berndverst:
# Description
Adds support for Actor Reentrancy

## Issue reference
https://github.com/dapr/python-sdk/issues/214
https://github.com/dapr/dapr/issues/2901

Please reference the issue this PR will close: #214

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


berndverst:
Hold off on reviewing. Will add some tests. Also it appears an existing GRPC test (which should be unrelated) is failing.

wcs1only:
> Hold off on reviewing. Will add some tests. Also it appears an existing GRPC test (which should be unrelated) is failing.

Forgive me if you know this already: You can actually denote a PR as a 'draft' to indicate that it isn't ready to be reviewed/merged. When you go to create a new PR, there's a little dropdown on the right side of the submit button that will let you 'Create as draft'

berndverst:
@wcs1only I think this is ready for review now.

By the way, how exactly are the samples run? Do you think I should add a sample to this PR or do that in a separate PR?

dapr/clients/http/dapr_actor_http_client.py
```diff
@@ -76,7 +85,7 @@ async def get_state(
             bytes: the value of the state.
         """
         url = f'{self._get_base_url(actor_type, actor_id)}/state/{name}'
-        body, r = await self._client.send_bytes(method='GET', url=url, data=None)
+        body, _ = await self._client.send_bytes(method='GET', url=url, data=None)
```
berndverst, line 88: There was an unused variable.

berndverst:
@halspang can you verify this looks good as far as the actors functionality goes (the right things are sent in the right spots)

dapr/actor/runtime/config.py
```diff
@@ -9,6 +9,31 @@
 from typing import Any, Dict, List, Optional
 
 
+class ActorReentrancyConfig:
+    def __init__(
+            self,
+            enabled: Optional[bool] = False,
```
halspang, line 15: While the reentrancy config itself is optional, this isn't an optional field in the config. How does this affect the actual serialization?

dapr/clients/http/dapr_actor_http_client.py
```diff
@@ -47,7 +49,14 @@ async def invoke_method(
         """
         url = f'{self._get_base_url(actor_type, actor_id)}/method/{method}'
 
-        body, r = await self._client.send_bytes(method='POST', url=url, data=data)
+        # import to avoid circular dependency
+        from dapr.actor.runtime.reentrancy_context import reentrancy_ctx
```
halspang, line 53: Can you explain how this makes a circular dependency?

dapr/clients/http/dapr_actor_http_client.py
```diff
@@ -47,7 +49,14 @@ async def invoke_method(
...
+        reentrancy_id = reentrancy_ctx.get()
+        headers: Dict[str, Union[bytes, str]] = (
+            {DAPR_REENTRANCY_ID_HEADER: reentrancy_id} if reentrancy_id else {})
```
halspang, line 56: Can this erase other headers that were already set/needed? Or is this the only header? I'm thinking of things like content-type or length.

tests/actor/test_actor_reentrancy.py
```diff
@@ -0,0 +1,223 @@
...
+        request_body = {
+            "message": "hello dapr",
+        }
+
+        reentrancy_id = "0faa4c8b-f53a-4dff-9a9d-c50205035085"
+
+        test_request_body = self._serializer.serialize(request_body)
+        response = _run(ActorRuntime.dispatch(
+            FakeMultiInterfacesActor.__name__, 'test-id',
+            "ReentrantMethod", test_request_body, reentrancy_id=reentrancy_id))
```
halspang, line 48: To confirm, this is just validating that we can call a reentrant actor normally right? Not actually testing the reentrant portion?

tests/actor/test_actor_reentrancy.py
```diff
@@ -0,0 +1,223 @@
...
+
+        async def dispatchReentrantCall(actorName: str, method: str, reentrancy_id: str):
+            return await ActorRuntime.dispatch(
+                actorName, 'test-id', method, request_body, reentrancy_id=reentrancy_id)
+
+        async def run_parallel_actors():
+            slow = dispatchReentrantCall(
+                FakeSlowReentrantActor.__name__, "ReentrantMethod", slow_reentrancy_id)
+            normal = dispatchReentrantCall(
+                FakeReentrantActor.__name__, "ReentrantMethod", normal_reentrancy_id)
```
halspang, line 77: Should this be the passthrough method? That's the only time the reentrant ID is fetched from the ContextVar.

halspang (commented):
Overall looks good. Just a few comments. For the header/reentrancy itself I think we're good aside from the config parsing.

dapr/clients/http/dapr_actor_http_client.py
berndverst: I honestly can't. When the import was at the module / library level all kinds of tests failed because the wrong things got imported. Doing this import here fixed the issue. The same thing is done in other places in the Python SDK. Basically what is described here: https://stackabuse.com/python-circular-imports

dapr/actor/runtime/config.py
berndverst: This does the right thing because of the default value, but the Optional qualifier is unnecessary. I'll remove this.

tests/actor/test_actor_reentrancy.py
berndverst: This simulates that a reentrant can be called normally way (as the final call in a reentrancy chain for example) but it only tests the internal part where the HTTP library has already parsed the reentrancy header and then calls the runtime dispatch function shown here. This particular test doesn't test calling a reentrant actor from a reentrant actor. That's done in another test.

tests/actor/test_actor_reentrancy.py
berndverst: I don't understand this comment. Could you explain @halspang?

This is just a bunch of Python weirdness to run two async methods truly in parallel and gather the results to force the interleaving of the processing. It is a sanity check to make sure the ContextVars truly work as we expected, that is, ContextVars are unique to each AsyncTask and are always copied forward into nested AsyncTasks.

I understand this looks weird - @wcs1only and I paired on this.

dapr/clients/http/dapr_actor_http_client.py
berndverst: Previously no custom headers were sent at all. That hasn't been supported with Actors in Python. The `invoke_method` only supports a single payload that is byte-encoded. This also only returns bytes.

The behavior prior to this PR is as follows:
Invoke_method always sends an empty dictionary for the headers via `send_bytes()`. Then that function checks whether the Content Type and DAPR API Token headers are presented (which they aren't) and if not uses defaults. See the `send_bytes` method definition and the test modification I made.

This introduces no breaking change.

berndverst:
Thanks for the review @halspang. Made a small change (removing the Optional qualifier which wasn't effective anyway). Otherwise take a look at my reply to your comments. Hope it all makes sense :)

tests/actor/test_actor_reentrancy.py
halspang: I'm fine with the python itself, my comment was more on how we'd handle the ContextVar in the code. Right now, this path only exercises the setting of the reentrant ID. If we use the passthrough method, it should also have to fetch the reentrant ID from the ContextVar. That's the path I'd like to check with interleaving.

halspang: @berndverst what do you think about my previous comment?

halspang (approved):
Spoke over a call about my last comment and I think what I'm looking for is more of an IT or auto-example. Once we have a good idea on what to make the actor do, we can add that in but we don't need to in this PR.

dapr/clients/http/dapr_actor_http_client.py
wcs1only: IIRC, the actual circular dependency was on dapr.actor.runtime which in turn depends on the http client.

wcs1only (approved):
Looks good. +1 for adding an example that exercises this.

PR #228, opened 2021-06-29
Enable dependabot

berndverst:
# Description

This enables dependabot to ensure we do not use vulnerable dependencies.

As of April 2021 dependabot can also scan dependencies defined in `setup.cfg`.


berndverst:
@wcs1only could you review this please?

wcs1only (approved):

PR #266, opened 2021-08-30
Adding optional Rule to subscription annotations for PubSub routing

pkedy:
# Description

This PR allows the Python SDK to make use of the [experimental PubSub feature](https://github.com/dapr/dapr/pull/3406) in Dapr 1.4. This is done by using a new, optional `rule` parameter of the `subscribe` annotation.

Example:

```python
# Specific handler using Pub/Sub routing
@app.subscribe(pubsub_name='pubsub', topic='TOPIC_A',
               rule=Rule("event.type == \"important\"", 1))
def mytopic_important(event: v1.Event) -> None:
    data = json.loads(event.Data())
    print(f'Received: id={data["id"]}, message="{data ["message"]}"' 
          ' content_type="{event.content_type}"',flush=True)
```

Note: gRPC/Protobufs were regenerated.

## Issue reference

Part of dapr/dapr#2582

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

/cc @wcs1only


ext/dapr-ext-grpc/dapr/ext/grpc/_servicier.py
```diff
@@ -53,19 +68,44 @@ def register_topic(
...
+
+        sub = registered_topic.subscription
+        rules = registered_topic.rules
+
+        if rule:
+            path = getattr(cb, '__name__', rule.match)
+            rules.append((rule.priority, appcallback_v1.TopicRule(
+                match=rule.match, path=path)))
+            rules.sort(key=lambda x: x[0])
+            rs = list(map(lambda r: r[1], rules))
```
wcs1only, line 106: So, looks like you're making a list out for the second value in the list of tuple pairs? Python has a concept called list comprehension for this:
```suggestion
            rs = [rule for id, rule in rules]
```

For another example:
```python
% python
Python 3.9.5 (default, May  4 2021, 03:33:11) 
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> pairs = [(1,"a"), (2, "b")]
>>> letters = [letter for number, letter in pairs]
>>> print(letters)
['a', 'b']
```
I think this is what you want?

ext/dapr-ext-grpc/dapr/ext/grpc/_servicier.py
```diff
@@ -53,19 +68,44 @@ def register_topic(
...
+            del sub.routes.rules[:]
+            sub.routes.rules.extend(rs)
```
wcs1only, line 108: Should be able to just set rules like this:
```python
sub.routes.rules = rs
```

Was there are reason to empty the list and then add everything back, instead of just replacing the existing list?

pkedy: Apparently, for repeating fields in proto, you cannot directly assign them or you get this error

image

https://stackoverflow.com/questions/23726335/how-to-assign-to-repeated-field

wcs1only: Ah, a protobufism, of course.

wcs1only (approved):

PR #273, opened 2021-09-08
Only upload builds to PyPI once

berndverst:
# Description

Only upload builds to PyPI once


wcs1only (approved):
Thanks for fixing this!

wcs1only (approved):

PR #276, opened 2021-09-10
Remove explicit protobuf dep from example

berndverst:
# Description

Removing the explicit protobuf dependency to avoid version mismatches. Dapr already provides the protobuf library.


mukundansundar (approved):
lgtm

Issue #295, opened 2021-11-09
[Question] Asyncio event loop in DaprInvocationHttpClient


wimvc1:
Hi all,

I have a question/remark regarding the `DaprInvocationHttpClient` and specifically the `invoke_method` method.
Source: https://github.com/dapr/python-sdk/blob/18a609ac01b6b3c91f0a9e736dc24c4fa4ed9ac3/dapr/clients/http/dapr_invocation_http_client.py

Starting from line 96:
```
loop = asyncio.new_event_loop()
asyncio.set_event_loop(loop)

return loop.run_until_complete(make_request())
```

This creates a new event loop which seems to cause issues when using other libraries/frameworks which also use asyncio.
More specifically:
```
RuntimeWarning: coroutine 'DaprInvocationHttpClient.invoke_method.<locals>.make_request' was never awaited
Cannot run the event loop while another loop is running
```

To my understanding, there should be some extra logic to check if there is an existing event loop. If so, reuse that loop to handle the `make_request()` call.

Looking forward to your comments/suggestions.

PS:
When using `nest-asyncio`, this issues seems to be resolved but we would like to minimise the number of dependencies in our application.
https://pypi.org/project/nest-asyncio/

wmeints:
Sounds like this could use an additional check. Async is always tricky.

i-am-shodan:
This is still important

Issue #90, opened 2020-07-10
Add support for async gRPC in Dapr SDK


youngbupark:
## Describe the proposal

> grpc_asyncio is still WIP. We may need to support it later when grpc_asyncio is stable.
> https://grpc.github.io/grpc/python/grpc_asyncio.html

Update: `grpc_asyncio` is now available, however we should publish an asynchronous Dapr Python SDK as a separate SDK, or at least sub package - like `dapr.aio` (`from dapr.aio.clients import DaprClient`)

There were concerns from several users about the performance overhead of calling asynchronous methods and running them in a synchronous context, so simply switching over to `grpc.aio` is not an option unfortunately.

XavierGeerinck:
Awesome proposal @youngbupark ! Really loving all the async await implementation work, it makes everything so much easier!

One question from my side: I see a lot of libraries that did not implement async / await yet... currently I need to use a patched asyncio with `run_until_complete` (e.g. `asyncio.get_event_loop().run_until_complete(self.proxy.SimActionSample())`) to make calls synchronous for these libraries to work with (Python and AI...).

Might it be interesting to introduce this to the SDK as well, or is it more interesting to have it in the documentation somewhere?

omidb:
Thanks for the proposal, this will unify a lot of things on the Python SDK. Any update on this @youngbupark ?appreciate it.

CodeMonkeyLeet:
I don't think @youngbupark is looking at these issues any more ... @wcs1only @artursouza for their opinion instead.

artursouza:
This work seems to be about using grpc_asyncio for all gRPC communication in the SDK.

omidb:
right now, it's really not easy to write an app to connect and interact with Actors or invoke a method on service or binding. Some are using async and some are not.  It would be great to see an example app that does multiple things such as communication with actors, using pubsub and invoke methods to examine the opportunities to make Python sdk more streamlined.

avvijeet:
Interested in working on this issue

berndverst:
Proposal:

Create a copy of the Dapr gRPC clients (`dapr.clients.grpc`) within a new package `dapr.aio.clients` and migrate these to `grpc.aio`.

Also create matching unit tests.

Ensure existing integration tests pass when they are updated to import `from dapr.aio.clients improt DaprClient` instead. This can be verified with `tox -e examples`

tmacam:
[...]

PR #300, opened 2021-11-14
Feature/http client invoke method async

wimvdc:
# Description

This is not an ideal solution. I haven't found a working solution to return the "awaited" result when a loop is already running.
I have added a new optional parameter to `invoke_method` named `return_coroutine` which allows the method to return a coroutine.

Example usage:
```
with DaprClient() as dapr_client:
            result = await dapr_client.invoke_method(app_id="dummyapp",
                method_name="my_dummy",
                data=None,
                return_coroutine=True)
```

Please advice on this approach or suggest a better one.

Note: Might be related to #90 which tries to do something similar for the grpc side of the client.

## Issue reference

Please reference the issue this PR will close: #295 

## Checklist

* [ ] Code compiles correctly
* [x] Created/updated tests


dapr/clients/http/dapr_invocation_http_client.py
```diff
@@ -39,7 +39,8 @@ def invoke_method(
             content_type: Optional[str] = None,
             metadata: Optional[MetadataTuple] = None,
             http_verb: Optional[str] = None,
-            http_querystring: Optional[MetadataTuple] = None) -> InvokeMethodResponse:
```
wcs1only, line 42: How about this? Can you just add an invoke_method_async call? I think having separate synchronous and async signatures might avoid confusion and maintain backwards compatibility.

wimvdc: Thanks for the response. This does indeed sound like a better approach.
Can you review this updated PR?

wimvdc:
Still waiting for some feedback

dapr/clients/__init__.py
```diff
@@ -115,3 +115,47 @@ def invoke_method(
...
+                app_id,
+                method_name,
+                data,
+                content_type=content_type,
+                metadata=metadata,
+                http_verb=http_verb,
+                http_querystring=http_querystring,
+                return_coroutine=True)
+        else:
+            return super().invoke_method(
```
paulyuk, line 154: Should this else case also be async by adding await?

e.g.
`return await super().invoke_method(

paulyuk (commented):
Like the new design of asyncInvoke functions.  See comment with q on super() case.

dapr/clients/__init__.py
paulyuk:
```suggestion
            return await super().invoke_method(
```

wcs1only:
My apologies, I tend to mentally filter out draft PRs when processing the backlog. I like your new approach. Do you consider this PR ready to be merged? If so, can you please mark it as "Ready for review" ?

dapr/clients/__init__.py
wimvdc: Firstly, thanks for the review and to answer your question/remark; no, I don't think so.
`super()` refers to `DaprGrpcClient`, the `invoke_method` there is completely synchronous. So as far my python knowledge goes, `await` it not needed here.

wimvdc:
Fixed the linter issue which was reported (all linter issue should be resolved now)

berndverst:
@wimvdc can you fix up the PR? The type check is failing

```
dapr/clients/__init__.py:151: error: Incompatible types in "await" (actual type "InvokeMethodResponse", expected type "Awaitable[Any]")  [misc]
dapr/clients/__init__.py:161: error: Incompatible return value type (got "InvokeMethodResponse", expected "Awaitable[InvokeMethodResponse]")  [return-value]
```

You can run this test with `tox -e mypy`

dapr/clients/__init__.py
```diff
@@ -106,20 +106,70 @@ def invoke_method(
...
+            content_type (str, optional): Content type of the data.
+            metadata (MetadataTuple, optional): Additional metadata or headers.
+            http_verb (str, optional): HTTP verb for the request.
+            http_querystring (MetadataTuple, optional): Query parameters.
+
+        Returns:
+            Union[InvokeMethodResponse,Awaitable[Any]]: the response from the method invocation.
+        """
+        if self.invocation_client:
+            result = self.invocation_client.invoke_method(
```
jfilipe-trd, line 157: Shouldn't be `result = await self.invocation_client.invoke_method` ? Otherwise the coroutine is never awaited.

berndverst:
Now it's the linter
```bash
tox -e flake8
```

```
./dapr/clients/__init__.py:131:17: F541 f-string is missing placeholders
./dapr/clients/__init__.py:141:101: E501 line too long (109 > 100 characters)
./dapr/clients/http/dapr_invocation_http_client.py:51:74: E231 missing whitespace after ','
```

dapr/clients/__init__.py
berndverst:
```suggestion
            result = await self.invocation_client.invoke_method(
```

berndverst (changes requested):
Please fix the linter warnings and take a look at the comment about awaiting the coroutine.

berndverst (dismissed):

berndverst (changes requested):
We cannot await a function that returns anything other than `Awaitable[Any]`

The mixed type approach doesn't work with our type hints. Can you find a way around that @wimvdc ?

wimvdc:
> We cannot await a function that returns anything other than `Awaitable[Any]`
> 
> The mixed type approach doesn't work with our type hints. Can you find a way around that @wimvdc ?

I can try.
I'm currently involved with some other stuff so it might take a while until I have some free time to look at this.

berndverst:
@wimvdc I rearchitected this PR quite a bit - but still keeping the essence of what you wanted. I really wanted to make sure your PR can make it into the the next release of the Python SDK!

berndverst (approved):

PR #302, opened 2021-11-15
Adding the ability to add an existing gRPC service

### LINKED ISSUE #301: Adding the capability to mix current gRPC services with the Python SDK gRPC extention
## Describe the proposal
If we can add a current service to the Python SDK gRPC extension server, then it will enable us to easily use gRPC proxying and also use other features that the SDK provides such as pub/sub etc.

omid-verneek:
# Description

I have added `add_external_service` to the gRPC extension. I also provided an example to showcase the feature. 
This PR helps our org to easily use our gRPC based services with the Python SDK. I think it can also help others to use gRPC proxying for invoking their methods. At the same time, the user gets the full capabilities that come with the SDK.

## Issue reference

This will close #301 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


ext/dapr-ext-grpc/dapr/ext/grpc/app.py
```diff
@@ -41,6 +41,10 @@ def __init__(self, **kwargs):
 
     def __del__(self):
         self.stop()
+    
+    def add_external_service(self, servicer_callback, external_servicer):
+        """Adds an external gRPC service to the same server"""
+        servicer_callback(external_servicer, self._server)
```
halspang, line 47: Is this for hosting the gRPC service?

examples/grpc_proxying/invoke-reciever.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+# async def serve() -> None:
+#     server = grpc.aio.server()
+#     helloworld_service_pb2_grpc.add_HelloWorldServiceServicer_to_server(HelloWorldService(), server)
+#     listen_addr = '127.0.0.1:50051'
+#     server.add_insecure_port(listen_addr)
+#     logging.info("Starting server on this: %s", listen_addr)
+#     await server.start()
+#     await server.wait_for_termination()
```
halspang, line 26: Commented code?

examples/grpc_proxying/invoke-caller.py
```diff
@@ -0,0 +1,45 @@
...
+
+import grpc
+import helloworld_service_pb2_grpc
+from helloworld_service_pb2 import HelloRequest, HelloReply
+from dapr.clients import DaprClient
+import json, time
+
+async def run() -> None:
+    async with grpc.aio.insecure_channel('127.0.0.1:50007') as channel:
+        metadata = (('dapr-app-id', 'server'),)
```
halspang, line 12: Will this be necessary for every call that the user makes to their gRPC service? Is it possible to provide this automatically? Perhaps when you construct the client?

halspang (commented):
Thanks for your submission! 

My knowledge of python is a little limited with gRPC, but the main comment I have is if we can provide the `dapr-app-id` header automatically when making calls. Requiring the user to provide it makes it feel a little clunky.

Here's what the .NET SDK did for reference: https://github.com/dapr/dotnet-sdk/pull/755

ext/dapr-ext-grpc/dapr/ext/grpc/app.py
omid-verneek: yes

examples/grpc_proxying/invoke-caller.py
omid-verneek: So the idea here is to follow gRPC proxying: [You can see here](https://docs.dapr.io/developing-applications/building-blocks/service-invocation/howto-invoke-services-grpc/)
This way a client can just use their grpc client to call the Dapr sidecar without using the client directly and Dapr will redirect it to the right service. All only by adding metadata.

examples/grpc_proxying/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+```bash
+pip3 install dapr dapr-ext-grpc
+```
+
+## Running in self-hosted mode
+
+Run the following command in a terminal/command-prompt:
+
+<!-- STEP -->
```
wcs1only, line 24: So these step annotation comments are not just for show, we actually have some automation which validates our examples with every run. It is capable of running your \`\`\`bash ... \`\`\` codeblocks and checking stdout. This avoids the somewhat tedious process of manually running the examples to make sure they work with every release.

Here's an example of what this looks like:
https://github.com/dapr/python-sdk/blame/master/examples/invoke-binding/README.md#L41-L47

Would you be willing to add these steps for at least the standalone (non-k8s) portion of your examples? You don't have to validate all of the output of your example, just enough to say "this example is working".

Here's the detailed guide for our example automation tool:
https://github.com/dapr/mechanical-markdown/tree/main/examples

Also, if you add your new example here, it will get run on every PR:
https://github.com/dapr/python-sdk/blob/master/tox.ini#L36-L45

To run validation locally:
```
pip3 install mechanical-markdown

cd examples/grpc_proxying
mm.py -l README.md
```

wcs1only (commented):
You have one minor lint error:
https://github.com/dapr/python-sdk/runs/4241882285?check_suite_focus=true#step:5:118

If you'd like to run the linter locally without waiting for our CI/CD, you can just run:
```
tox -e flake8
```

examples/grpc_proxying/README.md
omid-verneek: let me give it a shot

omid-verneek: @wcs1only should be good now

examples/grpc_proxying/helloworld_service_pb2_grpc.py
```diff
@@ -0,0 +1,70 @@
...
+            target,
+            options=(),
+            channel_credentials=None,
+            call_credentials=None,
+            insecure=False,
+            compression=None,
+            wait_for_ready=None,
+            timeout=None,
+            metadata=None):
+        return grpc.experimental.unary_unary(request, target, '/HelloWorldService/SayHello',
```
berndverst, line 66: I'm a bit concerned about this breaking our automated tests -- specifically the `validate examples` workflow we run via Mechnical Markdown.

We will certainly be updating the GRPC library as new versions come out. I assume given the experimental nature of this feature the maintainers of the GRPC library may be moving / removing this feature in the future.

Is there a different way to accomplish what you need to accomplish without the use of an experimental API?

omid-verneek: This is what is currently being generated using gRPC tools. I don't have control over this part. I have to mention this is part of the example and not part of the gRPC extension.

wcs1only: Yeah, this is something internal to the generated code. The main dapr API's grpc generated code has this too.

wcs1only (approved):
Awesome, thanks for your contribution.

PR #325, opened 2022-01-07
Adds support for State Store Query Alpha API

### LINKED ISSUE #296: add support for statestore query API
This feature is implemented in dapr runtime: dapr/dapr#3662

It should be available via the SDK

berndverst:
# Description

Adds support for State Store Query Alpha API

## Issue reference

Closes #296 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


berndverst:
@dmitsh can you review to make sure this looks good from a Query API perspective? I followed the Go SDK implementation closely.

berndverst:
FYI @wcs1only  -- validation tests are expected to fail in GitHub Actions because the 1.5.1 sidecar does not return the correct data. The sidecar needs to be manually built from master at this time.

dmitsh (approved):

wcs1only:
> FYI @wcs1only -- validation tests are expected to fail in GitHub Actions because the 1.5.1 sidecar does not return the correct data. The sidecar needs to be manually built from master at this time. This is expected to work once the 1.6.0 rc is live.

This might be a good case for having beta builds.

berndverst:
FYI nothing changed here -- I just rebased against master so that the Protobuff changes were no longer included in this PR

dapr/clients/grpc/_response.py
```diff
@@ -469,6 +469,110 @@ def items(self) -> Sequence[BulkStateItem]:
...
+    def __init__(
+            self,
+            key: str,
+            value: bytes,
+            etag: str = '',
+            error: str = ''):
+        """Initializes QueryResponseItem item from :obj:`runtime_v1.QueryStateItem`.
+
+        Args:
+            key (str): query reponse item's key.
```
paulyuk, line 491:
```suggestion
            key (str): query response item's key.
```

dapr/clients/grpc/_response.py
```diff
@@ -469,6 +469,110 @@ def items(self) -> Sequence[BulkStateItem]:
...
+            value (bytes): query reponse item's data.
```
paulyuk, line 492:
```suggestion
            value (bytes): query response item's data.
```

dapr/clients/grpc/_response.py
```diff
@@ -469,6 +469,110 @@ def items(self) -> Sequence[BulkStateItem]:
...
+            etag (str): query reponse item's etag.
```
paulyuk, line 493:
```suggestion
            etag (str): query response item's etag.
```

dapr/clients/grpc/_response.py
```diff
@@ -469,6 +469,110 @@ def items(self) -> Sequence[BulkStateItem]:
...
+
+
+class QueryResponse(DaprResponse):
+    """The response of state store query API.
+
+    This inherits from DaprResponse
+
+    Attributes:
+        results (Sequence[QueryResponseItem]): the query results.
+        token (str): query reponse token for pagination.
```
paulyuk, line 537:
```suggestion
        token (str): query response token for pagination.
```

dapr/clients/grpc/_response.py
```diff
@@ -469,6 +469,110 @@ def items(self) -> Sequence[BulkStateItem]:
...
+            self,
+            results: Sequence[QueryResponseItem],
+            token: str = '',
+            metadata: Dict[str, str] = dict(),
+            headers: MetadataTuple = ()):
+        """Initializes QueryResponse from :obj:`runtime_v1.QueryStateResponse`.
+
+        Args:
+            results (Sequence[QueryResponseItem]): the query results.
+            token (str): query reponse token for pagination.
```
paulyuk, line 551:
```suggestion
            token (str): query response token for pagination.
```

dapr/clients/grpc/client.py
```diff
@@ -461,6 +462,73 @@ def get_bulk_state(
...
+                pagination token and results of the query
+        """
+
+        if not store_name or len(store_name) == 0 or len(store_name.strip()) == 0:
+            raise ValueError("State store name cannot be empty")
+        req = api_v1.QueryStateRequest(
+            store_name=store_name,
+            query=query,
+            metadata=states_metadata)
+        response, call = self._stub.QueryStateAlpha1.with_call(req)
```
paulyuk, line 514: Q do we ever do async await patterns for SDK calls that could take a while?

paulyuk: Actually that seems out of scope but requested here: https://github.com/dapr/python-sdk/issues/90

paulyuk (commented):
Looks like a very thorough add of the feature plus example plus tests.

dapr/clients/grpc/client.py
berndverst: I'm keeping this consistent for now. Making everything async/await will be a different effort. It also makes testing a little harder :)

dapr/clients/grpc/client.py
```diff
@@ -459,6 +460,73 @@ def get_bulk_state(
             items=items,
             headers=call.initial_metadata())
 
+    def query_state_alpha1(
```
wcs1only, line 463: We had a chat about this offline. By appending alpha1(), we're basically forcing the user to do a code migration when this feature goes out of preview. In the ideal case, we'd hope that there aren't breaking changes from alpha1 -> release. In this scenario we'd have unnecessarily forced early adopters to change their code once the feature became release.

I propose we call the interface query_state, and simply include a warn() that this is an alpha feature, which we can simply remove when the final API goes to release, provided there are no changes

dmitsh: I had voiced same concerns with regards to go-sdk and other SDKs, but unfortunately wasn't able to persuade the maintainers.

wcs1only (approved):
LGTM

Issue #324, opened 2022-01-04
Deprecate gRPC metadata API field


wcs1only:
## Describe the proposal

In the early days of the dapr-python SDK we exposed gRPC metadata as an option parameter on each API call. This was before we had created the concept of Dapr metadata, so now we have a hugely confusing collision in terms, and in our own documentation: (See #299, #317). As I understand it, the only reason for exposing gRPC metadata was to allow the API token header. We've created a generic interceptor for the API token header, so there is really no use case that we're aware of for gRPC metadata. 

I propose deprecating and removing this field from the API with deprecation in 1.5 and removal in 1.7.

paulyuk:
@wcs1only I'm happy to pick this up

paulyuk:
I reviewed the situation.  I agree with this.
- there is a real issue where this param gets used by mistake to pass Dapr metadata (e.g. TTL) and users report issues
- the main scenario for gRPC metadata is to path metadata for auth to be interpreted by the server.  In this case Dapr is the runtime and we cannot think of other metadata that needs to pass through.  
-
I'm just reviewing this with a few other heavy gRPC users.. and then will create PR.

sahibble:
Hi, I have been using the python SDK recently with Dapr version 1.5.1 and have noticed I am getting these metadata deprecation warnings despite not passing in a metadata argument to any of the DaprClient methods.

I have had a look and I believe that some of the methods such as invoke_binding have a default empty tuple argument being passed in. 
```python
    def invoke_binding(
            self,
            binding_name: str,
            operation: str,
            data: Union[bytes, str],
            binding_metadata: Dict[str, str] = {},
            metadata: Optional[MetadataTuple] = ()) -> BindingResponse:
```
However it checks to see if metadata is not None which fails when the metadata is its default argument. 
```python
        if metadata is not None:
            warn('metadata argument is deprecated. Dapr already intercepts API token headers '
                 'and this is not needed.', DeprecationWarning, stacklevel=2)
```
I propose changing the default argument of metadata to None which would also make it align with some of the other methods.

PR #335, opened 2022-01-12
Add deprecation warnings for unused gRPC metadata

paulyuk:
# Description

In this release we will mark the optional metadata argument as deprecated in all gRPC client method calls.  We do this by raising a warning and updating the documentation.  In several releases the argument will be removed full stop.  

## Issue reference

Please reference the issue this PR will close: #324 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

Tests pass and now reflect the deprecation warning, which is expected until we remove the argument.

### DIFF
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -55,6 +55,7 @@ from dapr.clients.grpc._response import (
 
 from urllib.parse import urlencode
 
+from warnings import warn
 
 class DaprGrpcClient:
     """The convenient layer implementation of Dapr gRPC APIs.
@@ -158,9 +159,6 @@ class DaprGrpcClient:
                     method_name='method',
                     data=b'message',
                     content_type='text/plain',
-                    metadata=(
-                        ('header1', 'value1')
-                    ),
                 )
 
                 # resp.content includes the content in bytes.
@@ -178,9 +176,6 @@ class DaprGrpcClient:
                     app_id='callee',
                     method_name='method',
                     data=req_data,
-                    metadata=(
-                        ('header1', 'value1')
-                    ),
                 )
                 # Create protocol buffer object
                 resp_data = dapr_example_v1.CustomResponseMessage()
@@ -197,9 +192,6 @@ class DaprGrpcClient:
                     method_name='method',
                     data=b'message',
                     content_type='text/plain',
-                    metadata=(
-                        ('header1', 'value1')
-                    ),
                     http_verb='POST',
                     http_querystring=(
                         ('key1', 'value1')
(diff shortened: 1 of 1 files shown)


dapr/clients/grpc/client.py
```diff
@@ -215,13 +208,17 @@ def invoke_method(
             method (str): the method name which is called
             data (bytes or :obj:`google.protobuf.message.Message`): bytes or Message for data
                 which will send to id
-            metadata (tuple, optional): custom metadata
+            metadata (tuple, optional, DEPRECATED): custom metadata
```
wcs1only, line 211: So, for service invocation specifically, I think we  need to leave the metadata parameter. Service invocation acts as a pass through for gRPC requests to an arbitrary gRPC/HTTP endpoint running on another service. That arbitrary endpoint might care about what the headers/metadata are. So I think this is the one case where we need to preserve the ability to add arbitrary gRPC metadata. 

In other APIs, we're talking directly to the dapr endpoint, which does not use gRPC metadata, and instead uses a metadata field in the request body itself. Does this distinction make sense?

In the broader sense, we may want to be deprecating the concept of calling an invoke_method endpoint entirely in favor of gRPC/HTTP proxying which we now support. This is however, a much bigger discussion that is probably out of scope of the present effort.

dapr/clients/grpc/client.py
```diff
@@ -269,23 +266,23 @@ def invoke_binding(
...
                 # resp.data includes the response data in bytes.
-                # resp.metadata include the metadata returned from the external system.
 
         Args:
             binding_name (str): the name of the binding as defined in the components
             operation (str): the operation to perform on the binding
             data (bytes or str): bytes or str for data which will sent to the binding
             binding_metadata (dict, optional): metadata for output binding
-            metadata (tuple, optional): custom metadata to send to the binding
+            metadata (tuple, optional, DEPRECATED): custom metadata to send to the binding
```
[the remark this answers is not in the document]
paulyuk: Great idea.  I've updated each arg inline doc to clarify which arg is for Dapr metadata listened to by the component, and which one is the general gRPC.  

For example

`
Args:
            binding_name (str): the name of the binding as defined in the components
            operation (str): the operation to perform on the binding
            data (bytes or str): bytes or str for data which will sent to the binding
            binding_metadata (dict, optional): Dapr metadata for output binding
            metadata (tuple, optional, DEPRECATED): gRPC custom metadata
`

dapr/clients/grpc/client.py
paulyuk: This is a good concern.

I believe and lean to still deprecating this argument full stop, including for service invoke.  We still believe the use case is for informing the host of auth tokens.

I also agree we can't know what developers have already used gRPC metadata for in their own code.  Just for this scenario, and generally, I strongly recommend Dapr users adopt the gRPC and HTTP proxy approach to Service Invoke and do not use the SDK.  There will be much higher fidelity results.

For these reasons I recommend we move forward with the deprecation and monitor the situation after release and after the build warnings kick in.

wcs1only: I accept this reasoning.

wcs1only (approved):

PR #338, opened 2022-01-13
Fix incorrect license configuration

berndverst:

# Description

It needs to be `OSI Approved :: Apache Software License` not `Apache License` in the `setup.cfg`.

### DIFF
--- a/ext/dapr-ext-grpc/setup.cfg
+++ b/ext/dapr-ext-grpc/setup.cfg
@@ -7,7 +7,7 @@ license_file = LICENSE
 classifiers =
     Development Status :: 5 - Production/Stable
     Intended Audience :: Developers
-    License :: OSI Approved :: Apache License
+    License :: OSI Approved :: Apache Software License
     Operating System :: OS Independent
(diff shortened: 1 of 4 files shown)


ext/dapr-ext-grpc/setup.cfg
```diff
@@ -7,7 +7,7 @@ license_file = LICENSE
 classifiers =
     Development Status :: 5 - Production/Stable
     Intended Audience :: Developers
-    License :: OSI Approved :: Apache License
+    License :: OSI Approved :: Apache Software License
     Operating System :: OS Independent
     Programming Language :: Python
     Programming Language :: Python :: 3.7
```
wcs1only, line 13: While we're here, we should probably add 3.9 and 3.10

wcs1only (approved):

ext/dapr-ext-grpc/setup.cfg
berndverst: Done

wcs1only (approved):

Issue #336, opened 2022-01-12
How do you reject work with the GRPC ext client?


renatolfc:
In the [dapr-ext-grpc](https://github.com/dapr/python-sdk/tree/master/ext/dapr-ext-grpc) [documentation page](https://docs.dapr.io/developing-applications/sdks/python/python-sdk-extensions/python-grpc/), we see the following example for subscribing to `TOPIC_A` on the `pubsub` publish subscribe component.

```python
@app.subscribe(pubsub_name='pubsub', topic='TOPIC_A')
def mytopic(event: v1.Event) -> None:
    print(event.Data(),flush=True)
```

The document doesn't say how one `ACK`s a message. Are messages `ACK`ed when the decorated function (`mytopic`, in the example above) is called, or when it returns?

If a message is only `ACK`ed when the decorated function returns, how do I prevent redelivery when the decorated function needs to do a long-running operation? Or how do I reject work until a long-running operation completes?

What happens if an exception is raised from the decorated function? Will dapr try to redeliver the message?

Thanks!

renatolfc:
After some local experimentation, it seems that:

 1. The avoid retried deliveries, the decorated function should return as soon as it determines it can act upon a received event (seems to be similar to returning a `200` using the HTTP API)
 2. If an exception is raised from the decorated function, dapr will try to deliver the message at a later time (Seems to be similar to returning something `!= 200` in the HTTP API)

Is this description correct?

wcs1only:
One approach would be to allow the callback function to return a TopicEventResponse object.

renatolfc:
As a user, I like that suggestion. Should I try contributing a PR?

renatolfc:
#336 is my attempt at solving this issue following @wcs1only's suggestion.

Please let me know if it needs improvements, as it'd be super nice for me to be able to intruct dapr to retry/reject messages using the grpc client.

PR #341, opened 2022-01-14
Add topic event response support to GRPC ext client

rlfcunha:
# Description

This PR enables responding to topic events from the GRPC ext client. This allows telling dapr whether messages:

 1. Were received successfully
 2. Should be retried later
 3. Should be dropped

## Issue reference

Please reference the issue this PR will close: #336 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/examples/pubsub-simple/README.md
+++ b/examples/pubsub-simple/README.md
@@ -3,7 +3,7 @@
 This example utilizes a publisher and a subscriber to show the pubsub pattern, it also shows `PublishEvent`, `OnTopicEvent`, `GetTopicSubscriptions`, and `TopicEventResponse` functionality.
 It creates a publisher and calls the `publish_event` method in the `DaprClient`.
 It will create a gRPC subscriber and bind the `OnTopicEvent` method, which gets triggered after a message is published to the subscribed topic.
-The subscriber will tell dapr to retry delivery of the first message it receives, logging that the message will be retried, and printing it twice to standard output.
+The subscriber will tell dapr to retry delivery of the first message it receives, logging that the message will be retried, and printing it at least once to standard output.
 
 > **Note:** Make sure to use the latest proto bindings
 
--- a/examples/pubsub-simple/README.md
+++ b/examples/pubsub-simple/README.md
@@ -29,7 +29,6 @@ name: Run subscriber
 expected_stdout_lines:
   - '== APP == Subscriber received: id=1, message="hello world", content_type="application/json"'
   - 'RETRY status returned from app while processing pub/sub event'
-  - '== APP == Subscriber received: id=1, message="hello world", content_type="application/json"'
   - '== APP == Subscriber received: id=2, message="hello world", content_type="application/json"'
   - '== APP == Subscriber received: id=3, message="hello world", content_type="application/json"'
   - '== APP == Subscriber received: id=4, message="hello world", content_type="application/json"'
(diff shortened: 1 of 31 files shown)


paulyuk:
Thank you for this contribution @rlfcunha.

Is it possible to consider related docs, tests and examples updates at the same time?  E.g.

https://github.com/dapr/python-sdk/blob/master/daprdocs/content/en/python-sdk-docs/python-client.md

rlfcunha:
Sure @paulyuk. I've updated the PR. Please let me know if it needs further edits.

renatolfc:
@wcs1only, can you take a look and check whether this implements what you had suggested?

It'd be cool to be able to instruct the runtime on how to handle messages from the gRPC side of things...

Thanks!

berndverst:
This is the first SDK implementing this functionality @rlfcunha. Thanks for the contribution. As far as I can tell this is the right approach and the code looks good.

Could you also please modify the `pubsub-simple` example in the `examples` folder to verify returning the `retry` status actually redelivers the event and `success` status works as before?

That's the last piece missing -- this little end to end example that show this all actually work (especially because no SDK has implemented this yet). The `Readme.md` in that folder has special metadata that defines how the environment is spun up and the test is run. If you need more help with that syntax see https://github.com/dapr/mechanical-markdown

Once we have the `pubsub-simple` example updated this is ready to be merged! Thanks a lot!

berndverst (changes requested):
Please also modify the example in `examples/pubsub-simple` to show returning a retry status actually redelivers the same event.

rlfcunha:
Thanks for the feedback @berndverst. I'm updating the example to run the simplest case I could think of.

Please let me know whether that's ok.

examples/pubsub-simple/README.md
```diff
@@ -26,11 +27,14 @@ Run the following command in a terminal/command prompt:
 <!-- STEP
 name: Run subscriber
 expected_stdout_lines:
+  - '== APP == Subscriber received: id=1, message="hello world", content_type="application/json"'
+  - 'RETRY status returned from app while processing pub/sub event'
   - '== APP == Subscriber received: id=1, message="hello world", content_type="application/json"'
```
berndverst, line 32: This test is currently failing - therefore we see that this RETRY isn't actually working. ID=1 is not sent a second item.

rlfcunha: That's... interesting. I'm currently running my tests in a github codespaces instance, and tests pass there. Do you have any suggestions on how to better exercise this? Thanks.

yaron2:
> @rlfcunha I'm currently discussing the test failure with @yaron2. This may be expected -- apparently not every Dapr PubSub component implementation will automatically redeliver. In some cases redelivery will only occur after a new component instance connects or the existing component restarts. So maybe the sidecar needs to be restarted in this case.

Also, some implementations may take an arbitrary amount of time (seconds to minutes) before retrying, even if `daprd` does not need to restart and resubscribe.

examples/pubsub-simple/README.md
berndverst: How are you running this particular test? Also what version of `daprd` are you running this against?

berndverst, line 32: Maybe we can just remove checking for this second event and trust that it will be delivered eventually. This is assuming that you truly saw the redelivery work locally @rlfcunha.

Remove
```
- '== APP == Subscriber received: id=1, message="hello world", content_type="application/json"'
```

rlfcunha: As you've already concluded, the tests *are* non-deterministic.

Here's a screenshot showing expected messages matching:

retry "proof"

Now, when running the example tests with `tox -e examples`, I did see a test failure at `examples/invoke-custom-data`. Is it safe to assume this error is unrelated to this set of changes?

berndverst (approved):

PR #346, opened 2022-01-20
Fix actor example output

berndverst:

# Description

The `demo actor` example test using mechanical markdown appears to be broken because the test was originally written against incorrect output which has since been fixed in the sidecar.

This change should also be cherry picked into the release branch, but does not require a new RC.


berndverst:
Interestingly in CI this behaves differently from my local Mac.

berndverst:
Seems to be unnecessary -- likely caused by persistent test data locally

PR #351, opened 2022-01-28
Add PubSub subscriptions to Flask Dapr

berndverst:

# Description

Add PubSub subscriptions to Flask Dapr

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,11 +14,13 @@ commands =
     coverage run -m unittest discover -v ./tests
     coverage run -a -m unittest discover -v ./ext/dapr-ext-grpc/tests
     coverage run -a -m unittest discover -v ./ext/dapr-ext-fastapi/tests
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -14,11 +14,13 @@ commands =
     coverage run -m unittest discover -v ./tests
     coverage run -a -m unittest discover -v ./ext/dapr-ext-grpc/tests
     coverage run -a -m unittest discover -v ./ext/dapr-ext-fastapi/tests
+    coverage run -a -m unittest discover -v ./ext/flask_dapr/tests
```
artursouza, line 17: shouldn't we follow the dapr-ext-* pattern for this directory too?

berndverst: unfortunately the package has always been called `flask_dapr` whereas the others are `dapr-ext-grpc` and `dapr-ext-fastapi`

wcs1only (approved):

PR #356, opened 2022-02-15
Fixes #354 by changing default value of soon to be removed grpc metadata

### LINKED ISSUE #354: Metadata deprecation warning despite not passing in the argument
I have been using the python SDK recently with Dapr version 1.5.1 (and python version 3.9.6) and have noticed I am getting these metadata deprecation warnings despite not passing in a metadata argument to any of the DaprClient methods. I would expect there to be no warning if I leave out the metadata parameter. Behaviour may be after PR #335 and issue #324.

I have had a look and I believe that some of the methods such as invoke_binding have an empty tuple assigned in as a default parameter to metadata.

```python
    def invoke_binding(
            self,
            binding_name: str,
            operation: str,
            data: Union[bytes, str],
            binding_metadata: Dict[str, str] = {},
            metadata: Optional[MetadataTuple] = ()) -> BindingResponse:
```
However it checks to see if metadata is not None which fails when the metadata is its default argument (the empty tuple).
```python
        if metadata is not None:
            warn('metadata argument is deprecated. Dapr already intercepts API token headers '
                 'and this is not needed.', DeprecationWarning, stacklevel=2)
```
I propose changing the default argument of metadata to None which would also make it align with some of the other methods if this is not expected behaviour.

berndverst:

Fixes #354

Changing default value of soon to be removed grpc metadata

* [x] Code compiles correctly

### DIFF
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -127,7 +127,7 @@ class DaprGrpcClient:
 
     def _get_http_extension(
             self, http_verb: str,
-            http_querystring: Optional[MetadataTuple] = ()
+            http_querystring: Optional[MetadataTuple] = None
     ) -> common_v1.HTTPExtension:  # type: ignore
         verb = common_v1.HTTPExtension.Verb.Value(http_verb)  # type: ignore
         http_ext = common_v1.HTTPExtension(verb=verb)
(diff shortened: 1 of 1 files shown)


berndverst:
@sahibble this should address your issue report.

halspang (approved):

Issue #334, opened 2022-01-11
Add support for Configuration API (Subscribe Configuration API)


amulyavarote:
## Describe the proposal

The Configuration API feature has been implemented in Dapr runtime: dapr/dapr#3606
It needs to be accessible via the SDK. This issue is to add method for subscribe configuration API.

PR #369, opened 2022-03-16
Added subscribe config API support

amulyavarote:
Signed-off-by: Amulya Varote <amulyavarote@microsoft.com>

# Description

Added subscribe config API support.

## Issue reference

Please reference the issue this PR will close: #334 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


daprdocs/content/en/python-sdk-docs/python-client.md
```diff
@@ -166,6 +166,33 @@ with DaprClient() as d:
...
 ```
 
+### Subscribe to configuration
+
+```python
+import asyncio
+from time import sleep
+from dapr.clients import DaprClient
+
+async def executeConfiguration():
```
berndverst, line 176: Please use 4 spaces for indentation

berndverst (changes requested):
Please update examples/configuration/Readme.md to add a mechanical markdown test for the Subscribe.

You may need to bring up one client to watch for configuration changes, and another app to actually modify the configuration change to trigger this.

Also, please also fix the indentation on your existing `examples/configuration` file. The indentation of everything in that folder should use 4 spaces, not 8.

dapr/clients/grpc/_response.py
```diff
@@ -661,6 +664,31 @@ def items(self) -> Sequence[ConfigurationItem]:
...
+        return self.items
+
+    def watch_configuration(self, stub, store_name, keys, config_metadata):
+        req = api_v1.SubscribeConfigurationRequest(
+            store_name=store_name, keys=keys, metadata=config_metadata)
+        thread = threading.Thread(target=self.read_subscribe_config, args=(stub, req))
+        thread.daemon = True
+        thread.start()
+
+    def read_subscribe_config(self, stub, req):
```
mukundansundar, line 681: @berndverst Do you think this should be `_read_subscribe_config` just to let users know it is an internal method and should not be used outside the class?

mukundansundar: @berndverst thoughts on this?

dapr/clients/grpc/client.py
```diff
@@ -916,6 +917,40 @@ def get_configuration(
             items=items,
             headers=call.initial_metadata())
 
+    async def subscribe_configuration(
```
mukundansundar, line 920: I have a quick question ... Because of the async nature of the request this method is defined in the async await format ... But I find it odd that a Client class has all synchronous methods and the lone async method. .... 
cc @berndverst

dapr/clients/grpc/_response.py
berndverst: I agree. @amulyavarote, can you update that? Private / internal method should begin with an underscore `_`

dapr/clients/grpc/client.py
berndverst: Yes this is odd, but also presents our first async grpc API

We probably need to support this eventually:
https://github.com/dapr/python-sdk/issues/90

mukundansundar: That is true ...

berndverst (approved):

PR #379, opened 2022-04-01
Added right unit test cases

amulyavarote:
Signed-off-by: Amulya Varote <amulyavarote@microsoft.com>

# Description

Added right unit test cases

## Issue reference

Please reference the issue this PR will close: NA

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


tests/clients/test_dapr_grpc_client.py
```diff
@@ -23,17 +23,24 @@
...
 from .fake_dapr_server import FakeDaprSidecar
 from dapr.conf import settings
 from dapr.clients.grpc._helpers import to_bytes
 from dapr.clients.grpc._request import TransactionalStateOperation
 from dapr.clients.grpc._state import StateOptions, Consistency, Concurrency, StateItem
+from dapr.clients.grpc._response import (
+    ConfigurationItem
+)
+
+from tests.actor.utils import _run
```
berndverst, line 36: I know this is convenient, but please don't use that here because it's just strange to import from the actor specific test utilities.

berndverst:
This is not quite the right approach for this test @amulyavarote, but it is almost there.

DaprClient.subscribe_configuration() should return a partially fake (mocked) ConfigurationWatcher and then get_items() should return your desired data (that method can be mocked too).

berndverst:
You can so something like this... this is just a rough idea

```python
from mock import patch

def mock_watch(self, stub, store_name, keys, config_metadata):
    # we can directly manipulate our keys now
    for key in keys:
      self.items.append(ConfigurationItem(
          key=item.key,
          value="test",
          version="1.7.0"))

with patch.object(ConfigurationWatcher, 'watch_configuration', mock_watch):
    watcher = _run(dapr.subscribe_configuration(store_name="configurationstore", keys="k"))
    res = watcher.get_items() # this will already have data
    # check whether res[0]['key'] == 'k'
```

berndverst (approved):

PR #395, opened 2022-04-23
Fix flakey pubsub integration test

berndverst:

# Description

Fixes the pubsub example integration test which was extremely flakey. Also addressed some style issues.

### DIFF
--- a/examples/pubsub-simple/README.md
+++ b/examples/pubsub-simple/README.md
@@ -36,7 +36,7 @@ expected_stdout_lines:
   - '== APP == Wildcard-Subscriber received: id=6, message="hello world", content_type="application/json"'
 output_match_mode: substring
 background: true
-sleep: 5 
+sleep: 3 
 -->
 
 ```bash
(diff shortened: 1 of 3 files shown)


examples/pubsub-simple/README.md
```diff
@@ -75,14 +71,12 @@ dapr run --app-id python-publisher --app-protocol grpc --dapr-grpc-port=3500 pyt
 
 <!-- STEP
 expected_stdout_lines: 
-  - '✅  app stopped successfully: python-publisher'
```
artursouza, line 78: Why the publisher is not stopped anymore?

artursouza (changes requested):

examples/pubsub-simple/README.md
berndverst: Because we are publishing only a finite number of events and it exits itself. Previously we were publishing an infinite number.

artursouza (approved):

PR #404, opened 2022-05-18
Add FastAPI router tags for OpenAPI docs

### LINKED ISSUE #401: Passing APIRouter tags to the fast-api DaprActor
## Describe the proposal
The routes from the DaprActor appears in the openapi documentation so i think the library should allow to pass the router tags or supply default tags in order to differentiate the Actor routes from the application routes
The new Actor should look like:
```python
class DaprActor(object):
    def __init__(self, app: FastAPI, router_tags: List[str]): <----- New
        self._dapr_serializer = DefaultJSONSerializer()
        self._router = APIRouter(tags=tags) <----- New
        self.init_routes(self._router)
        app.include_router(self._router)
......
```
image
__Will now look like__
image

berndverst:

# Description

Implements default Flask API Router Tags and the ability to customize these for Dapr Actor endpoints and Dapr PubSub endpoints in the Dapr Fast API extension. This is only used when generating OpenAPI documentation. (See referenced issue)

## Issue reference

Closes #401

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
+++ b/ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
@@ -26,6 +26,9 @@ from dapr.serializers import DefaultJSONSerializer
 DEFAULT_CONTENT_TYPE = "application/json; utf-8"
 DAPR_REENTRANCY_ID_HEADER = 'Dapr-Reentrancy-Id'
 
+# This should be added to all magic Dapr Actor methods implemented here
+DEFAULT_ACTOR_ROUTER_TAGS: List[str] = ['Actor']
+
 
 def _wrap_response(
         status_code: int,
@@ -48,11 +51,10 @@ def _wrap_response(
 
 
 class DaprActor(object):
-    _DEFAULT_ACTOR_ROUTER_TAGS: List[str] = ['Actor']
 
     def __init__(self, app: FastAPI,
-                 router_tags: Optional[List[str]] = None):
-        self._router_tags = router_tags if router_tags else self._DEFAULT_ACTOR_ROUTER_TAGS
+                 router_tags: Optional[List[str]] = DEFAULT_ACTOR_ROUTER_TAGS):
+        self._router_tags = router_tags
         self._router = APIRouter()
         self._dapr_serializer = DefaultJSONSerializer()
         self.init_routes(self._router)
--- a/ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
+++ b/ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from typing import Any, Optional, Type
+from typing import Any, Optional, Type, List
 
 from fastapi import FastAPI, APIRouter, Request, Response, status   # type: ignore
 from fastapi.logger import logger
@@ -48,23 +48,27 @@ def _wrap_response(
 
 
 class DaprActor(object):
-    def __init__(self, app: FastAPI):
-        self._dapr_serializer = DefaultJSONSerializer()
+    _DEFAULT_ACTOR_ROUTER_TAGS: List[str] = ['Actor']
+
+    def __init__(self, app: FastAPI,
+                 router_tags: Optional[List[str]] = None):
+        self._router_tags = router_tags if router_tags else self._DEFAULT_ACTOR_ROUTER_TAGS
         self._router = APIRouter()
+        self._dapr_serializer = DefaultJSONSerializer()
         self.init_routes(self._router)
         app.include_router(self._router)
 
     def init_routes(self, router: APIRouter):
-        @router.get("/healthz")
+        @router.get("/healthz", tags=self._router_tags)
         async def healthz():
             return {'status': 'ok'}
 
-        @router.get('/dapr/config')
+        @router.get('/dapr/config', tags=self._router_tags)
         async def dapr_config():
             serialized = self._dapr_serializer.serialize(ActorRuntime.get_actor_config())
             return _wrap_response(status.HTTP_200_OK, serialized)
 
-        @router.delete('/actors/{actor_type_name}/{actor_id}')
+        @router.delete('/actors/{actor_type_name}/{actor_id}', tags=self._router_tags)
         async def actor_deactivation(actor_type_name: str, actor_id: str):
             try:
                 await ActorRuntime.deactivate(actor_type_name, actor_id)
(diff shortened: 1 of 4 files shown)


ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py
```diff
@@ -69,7 +74,8 @@ def decorator(func):
 
             self._app.add_api_route(event_handler_route,
                                     func,
-                                    methods=["POST"])
+                                    methods=["POST"],
```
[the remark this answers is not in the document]
berndverst: That is how Dapr PubSub works and this is merely a convenience method to provide an event handler compatible with the expected route path for the pubsub source and topic. The Sidecar only does POST requests for PubSub. This is *not* for service invocation or Actors.

ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
```diff
@@ -48,23 +48,27 @@ def _wrap_response(
...
+                 router_tags: Optional[List[str]] = None):
+        self._router_tags = router_tags if router_tags else self._DEFAULT_ACTOR_ROUTER_TAGS
         self._router = APIRouter()
+        self._dapr_serializer = DefaultJSONSerializer()
         self.init_routes(self._router)
         app.include_router(self._router)
 
     def init_routes(self, router: APIRouter):
-        @router.get("/healthz")
+        @router.get("/healthz", tags=self._router_tags)
```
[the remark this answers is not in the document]
berndverst: Sure - though I'm the only maintainer of the SDK so the risk someone forgets is slim :)

I think it goes back to the intent behind this. I figured you want the automatically provided routes to be annotated but anything you implement yourself doesn't need to be annotated.

I think it would be confusing to annotate user-provided handlers (even if they are on an actor). This is why I think only the "magic" methods should be annotated automatically.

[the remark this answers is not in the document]
yairsimantov20: Thats not what i meant
what i meant is when and if you will add another "magic" method we wouldn't want you to forget and add the tags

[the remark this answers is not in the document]
berndverst: Yes I understand that part - you want a comment so that a maintainer of this code will not forget to apply the tags to other magic methods.

ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
```diff
@@ -48,23 +48,27 @@ def _wrap_response(
 
 
 class DaprActor(object):
-    def __init__(self, app: FastAPI):
-        self._dapr_serializer = DefaultJSONSerializer()
+    _DEFAULT_ACTOR_ROUTER_TAGS: List[str] = ['Actor']
+
+    def __init__(self, app: FastAPI,
+                 router_tags: Optional[List[str]] = None):
+        self._router_tags = router_tags if router_tags else self._DEFAULT_ACTOR_ROUTER_TAGS
```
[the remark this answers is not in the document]
berndverst: Almost.

If you don't want router tags you need to specify `[]`. `None` means you need to use the default so this will be:

```python
self._router_tags = self._DEFAULT_ACTOR_ROUTER_TAGS if router_tags is None else router_tags
```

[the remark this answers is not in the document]
yairsimantov20: True 
Almost missed that ty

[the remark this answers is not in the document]
berndverst: I just changed this a little bit. You can provide either '[]' or 'None' to disable router tags going forward. Pushing that change now!

ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
```diff
@@ -26,6 +26,9 @@
 DEFAULT_CONTENT_TYPE = "application/json; utf-8"
 DAPR_REENTRANCY_ID_HEADER = 'Dapr-Reentrancy-Id'
 
+# This should be added to all magic Dapr Actor methods implemented here
+DEFAULT_ACTOR_ROUTER_TAGS: List[str] = ['Actor']
```
yairsimantov20, line 30: Don't you think that putting this inside the actor class will be more convenient if someone will want to inherit the class and change the default tags?

ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py
```diff
@@ -24,13 +27,16 @@ class DaprApp:
         app_instance: The FastAPI instance to wrap.
     """
 
-    def __init__(self, app_instance: FastAPI):
+    def __init__(self, app_instance: FastAPI,
+                 router_tags: Optional[List[str]] = DEFAULT_ROUTER_TAGS):
```
yairsimantov20, line 31: i don't recommend using the default value like this
`DEFAULT_ROUTER_TAGS` is a list and its immutable which can cause some miss understanding
For example if you will have the next code it can cause some issues
```
a = []
def c(b=a):
    return b
 
 print(a())   # []
 a().append(1)
 print(a()) # [1]
 ```
 first call will have an empty list and on the next one we will append value to the result list which will cause the next runs to have that value as well

Comment (yairsimantov20) [reply ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py:31]: this comment applies to actor.py as well

**Review by yairsimantov20 — CHANGES_REQUESTED** [review]
I think the previous solution was better

File: ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py (hunk 5 above)
Comment (artursouza) [reply ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py:31]: This is a great comment and I tried the code above and it would be a concern if router_tags would to be returned to the caller. On the other hand, it is kept private within the SDK code.

To avoid this, we could copy-and-paste `['xyz']` into each method but it would hurt readability IMO.

Comment (artursouza) [reply ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py:31]: I would recommend name the default as private with _ to communicate these should not be touched.

Comment (berndverst) [reply ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py]: My final offer 😄 

```python
    def __init__(self, app_instance: FastAPI,
                 router_tags: Optional[List[str]] = ['PubSub']):
```

ext/dapr-ext-fastapi/dapr/ext/fastapi/actor.py
berndverst: That's true. I changed the approach now.

As long as you call `super().__init__()` this should be fine now.

ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py
berndverst: By the way in your example @yairsimantov20 `a` was not intended to be mutable. It was a constant. Anyone who wanted to use a different value for `a` (`router_tags`) needs to provide it in the constructor as a keyword argument!

artursouza (approved):

Issue #410, opened 2022-06-01
Provide a support for Dapr sidecar Metadata API


pravinpushkar:
## Expected Behavior

SDK should provide a support for Sidecar Metadata APIs (HTTP and gRPC)- 
1. Get Metadata
2. Set Metadata

## Actual Behavior

It does not support Metadata APIs

## Release Note

RELEASE NOTE: **ADD** Support for Sidecar Metadata APIs

berndverst:
cc @tmacam also take a look at this

berndverst:
Please add mechanical markdown tests. I will merge this once Dapr/Dapr 1.9 RC.1 is out.

Issue #445, opened 2022-08-18
Update Metadata API example code to check for all fields in response object.


tmacam:
## Expected Behavior

The current GRPC implementation of the [Metadata API] in dapr 1.18 returns a response missing some fields. 

Tracking issues in dapr/dapr:
1. dapr/dapr#5051, 
2. dapr/dapr#5038 and 
3. dapr/dapr#5018.

In the current stable version of Dapr, only `extended_metadata` is returned correctly. This should be addressed in dapr/dapr itself but, once it is, we need to update our example for the Metadata API to check for these missing fields.

File `examples/metadata/app.py` should be updated to something like
```diff
--- a/examples/metadata/app.py
+++ b/examples/metadata/app.py
@@ -28,12 +28,10 @@ def main():

         print("Now, we will fetch the sidecar's metadata ")
         metadata = dapr.get_metadata()
+        print(f'{metadata.application_id=}"')
+        print(f'{metadata.active_actors_count=} should be non-empty')
+        print(f'{metadata.registered_components=} should be non-empty')
+        print(f'{metadata.extended_metadata=} should be non-empty')

         old_value = metadata.extended_metadata[extended_attribute_name]
```

and `examples/metadata/README.md` should be updated to something like

```diff
--- a/examples/metadata/README.md
+++ b/examples/metadata/README.md
@@ -34,6 +34,10 @@ name: Run state store example
 expected_stdout_lines:
   - "== APP == First, we will assign a new custom label to Dapr sidecar's"
   - "== APP == Now, we will fetch the sidecar's metadata"
+  - "== APP == metadata.application_id='' but should be "my-metadata-app""
+  - "== APP == metadata.active_actors_count={} but should be non-empty"
+  - "== APP == metadata.registered_components=[] but should be non-empty"
+  - "== APP == metadata.extended_metadata={'is-this-our-metadata-example': 'yes'}"
   - "== APP == We will update our custom label value and check it was persisted"
   - "== APP == We added a custom label named extended_attribute_name='is-this-our-metadata-example'"
   - "== APP == Its old value was old_value='yes' but now it is new_value='You bet it is!'"
@@ -50,6 +54,10 @@ The output should be as follows:
 == APP == First, we will assign a new custom label to Dapr sidecar's
 == APP == Now, we will fetch the sidecar's metadata
+== APP == metadata.application_id='' but should be "my-metadata-app"
+== APP == metadata.active_actors_count={} but should be non-empty
[...]

PR #446, opened 2022-08-18
Metadata API support for Python GRPC clients.

tmacam:
# Description

This PR adds support for [Metadata API] to the Python SDK, along with tests and a sample example code.

Two new methods were added to the python GRPC client:

1. **get_metadata**: Gets the Dapr sidecar information provided by
   the Metadata Endpoint.
2. **set_metadata**: Adds a custom label to the Dapr sidecar
   information stored by the Metadata endpoint.

It might be worth pointing out that while this PR is code-complete, the current GRPC implementation of the [Metadata API] in dapr 1.18 returns a response missing some fields. See dapr/dapr#5038 and dapr/dapr#5018. This is being tracked in #445.

[Metadata API]: https://docs.dapr.io/reference/api/metadata_api/

## Issue reference
Please reference the issue this PR will close: #410 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/dapr/clients/grpc/_response.py
+++ b/dapr/clients/grpc/_response.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import contextlib
 import threading
 from enum import Enum
-from typing import Dict, Optional, Union, Sequence, List, TYPE_CHECKING
+from typing import Dict, Optional, Union, Sequence, List, TYPE_CHECKING, NamedTuple
 
 from google.protobuf.any_pb2 import Any as GrpcAny
 from google.protobuf.message import Message as GrpcMessage
@@ -851,3 +851,69 @@ class TryLockResponse(contextlib.AbstractContextManager, DaprResponse):
         if self._success:
             self._client.unlock(self._store_name, self._resource_id, self._lock_owner)
         # else: there is no point unlocking a lock we did not acquire.
+
+
+class GetMetadataResponse(DaprResponse):
+    '''GetMetadataResponse is a message that is returned on GetMetadata rpc call.'''
+
+    def __init__(
+        self,
+        application_id: str,
+        active_actors_count: Dict[str, int],
+        registered_components: Sequence[RegisteredComponents],
+        extended_metadata: Dict[str, str],
+        headers: MetadataTuple = (),
+    ):
+        '''Initializes GetMetadataResponse.
+
+        Args:
+            application_id (str): The Application ID.
+            active_actors_count (Dict[str, int]): mapping from the type of
+                    registered actors to their number of running instances.
+            registered_components (Sequence[RegisteredComponents]): list of
+                    loaded components metadata.
+            extended_metadata (Dict[str, str]): mapping of custom (extended)
+                    attributes to their respective values.
+            headers (Tuple, optional): the headers from Dapr gRPC response.
+        '''
+        super().__init__(headers)
+        self._application_id = application_id
+        self._active_actors_count = active_actors_count
+        self._registered_components = registered_components
+        self._extended_metadata = extended_metadata
+
+    @property
+    def application_id(self) -> str:
+        '''The Application ID.'''
+        return self._application_id
+
+    @property
+    def active_actors_count(self) -> Dict[str, int]:
+        '''Mapping from the type of registered actors to their number of running instances.'''
+        return self._active_actors_count
+
+    @property
+    def registered_components(self) -> Sequence[RegisteredComponents]:
+        '''List of loaded components metadata.'''
+        return self._registered_components
+
+    @property
+    def extended_metadata(self) -> Dict[str, str]:
+        '''Mapping of custom (extended) attributes to their respective values.'''
+        return self._extended_metadata
+
+
+class RegisteredComponents(NamedTuple):
+    '''Describes a loaded Dapr component.'''
+
+    name: str
+    '''Name of the component.'''
+
+    type: str
+    '''Component type.'''
+
+    version: str
+    '''Component version.'''
+
+    capabilities: Sequence[str]
+    '''Supported capabilities for this component type and version.'''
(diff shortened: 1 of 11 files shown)


dapr/clients/grpc/_response.py
```diff
@@ -851,3 +851,70 @@ def __exit__(self, *exc) -> None:
...
+        '''Mapping from the type of registered actors to their number of running instances.'''
+        return self._active_actors_count
+
+    @property
+    def registered_components(self) -> Sequence[RegisteredComponents]:
+        '''List of loaded components metadata.'''
+        return self._registered_components
+
+    @property
+    def extended_metadata(self) -> Dict[str, str]:
```
tmacam, line 902: I sincerely considered renaming this to `custom_attributes` because it is less cryptic than `extended_metadata`.

But doing so would deviate a bit too much from the GRPC interface and I thought it would lead to client confusion.

examples/metadata/app.py
```diff
@@ -0,0 +1,42 @@
...
+        dapr.set_metadata(extended_attribute_name, 'You bet it is!')
+        metadata = dapr.get_metadata()
+        new_value = metadata.extended_metadata[extended_attribute_name]
+        print("We added a custom label named [%s]" % extended_attribute_name)
+        print("Its old value was [%s] but now it is [%s]" % (old_value, new_value))
+        print("And we are done 👋", flush=True)
+
+
+if __name__ == '__main__':
+    # Suppress "The Distributed Lock API is an Alpha" warnings
```
tmacam, line 41: Left over comment. Better remove it.

tests/clients/test_dapr_grpc_client.py
```diff
@@ -678,6 +678,75 @@ def test_unlock_input_validation(self):
...
+            for metadata_value in [str(i) for i in range(10)]:
+                dapr.set_metadata(attributeName=metadata_key,
+                                  attributeValue=metadata_value)
+                response = dapr.get_metadata()
+                self.assertIsNotNone(response)
+                self.assertIsNotNone(response.extended_metadata)
+                self.assertEqual(response.extended_metadata[metadata_key],
+                                 metadata_value)
+            # Empty string and blank strings should be accepted just fine
+            # By this API
```
tmacam, line 726: s/By/by/

dapr/clients/grpc/client.py
```diff
@@ -1120,6 +1123,66 @@ def wait(self, timeout_s: float):
...
+        The metadata API returns a list of the components loaded,
+        the activated actors (if present) and attributes with information
+        attached.
+
+        Each loaded component provides its name, type and version and also
+        information about supported features in the form of component
+        capabilities.
+        """
+        _resp, call = self._stub.GetMetadata.with_call(GrpcEmpty())
+        response: api_v1.GetMetadataResponse = _resp  # type alias
```
tmacam, line 1139: This was done just for the sake of having type-hinting on the `response` object. An alternative way, as discussed on https://stackoverflow.com/questions/52082939/type-hints-when-unpacking-a-tuple, is to just declare `response`  type directly before its usage:

```python
response: api_v1.GetMetadataResponse 
response, call = self._stub.GetMetadata.with_call(GrpcEmpty())
```

Same number of lines but one less variable ;)

mukundansundar: I think this seems better and recommended way for unpacking tuples ...

examples/metadata/app.py
```diff
@@ -0,0 +1,42 @@
...
+
+        print("We will update our custom label value and check it was persisted")
+        dapr.set_metadata(extended_attribute_name, 'You bet it is!')
+        metadata = dapr.get_metadata()
+        new_value = metadata.extended_metadata[extended_attribute_name]
+        print("We added a custom label named [%s]" % extended_attribute_name)
+        print("Its old value was [%s] but now it is [%s]" % (old_value, new_value))
+
+        print("And we are done 👋", flush=True)
+
```
mukundansundar, line 39: I think apart from getting and setting metadata extended attributes, we can also show what the value of the components metadata field is ...

examples/metadata/components/lockstore.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: lockstore
```
mukundansundar, line 4: why is this component needed in metadata example?

examples/metadata/app.py
tmacam: That was the plan but I am was not receiving components metadata on the GRPC answer. I want to debug this further but I don't want to have this dapr/dapr issue tangled with the Python SDK. I cut issue #445 to keep track of this.

examples/metadata/components/lockstore.yaml
tmacam: Not really needed. I thought the more the merrier but I guess the 2 "traditional" components (pubsub and state store on redis)  should be more than enough. I will remove this one.

berndverst: Any component that works with the standard Dapr installation is fine. I don't care which one we are using @mukundansundar

berndverst:
Can you please already make the changes you would be making if https://github.com/dapr/dapr/issues/5051 were completed and effectively address #445. I will merge this PR once dapr/dapr 1.9.0 rc.1 is out so I can run the tests in CI/CD. But in the meantime I will manually build the sidecar and use that once you completed the PR.

berndverst (approved):

tmacam:
> Can you please already make the changes you would be making if [dapr/dapr#5051](https://github.com/dapr/dapr/issues/5051) were completed and effectively address #445. I will merge this PR once dapr/dapr 1.9.0 rc.1 is out so I can run the tests in CI/CD. But in the meantime I will manually build the sidecar and use that once you completed the PR.
> 
> EDIT: Never mind, let's open a separate PR for this - but it needs to be ready by the time we have Dapr/Dapr 1.9.0 rc.1

Separate PR created: #450

PR #458, opened 2022-09-19
Add support for large messages in `dapr.ext.grpc`

### LINKED ISSUE #451: Ability to receive messages >4MB when subscribing to pubsub
## Summary

I have a scenario where a Python app subscribes to Dapr pubsub using `dapr.ext.grpc`'s `App`. When receiving messages >4MB I see the following error in the Dapr sidecar:

```text
error: code = ResourceExhausted desc = Received message larger than max (20000158 vs. 4194304)
```

I found that the ability to specify max messages size was added to `DaprClient` in #371.  Is there a way to increase the max message size to be able to receive messages >4MB with `App` subscriptions?

berndverst:

# Description

Adds option to specify the maximum gRPC message length in bytes for the `dapr.ext.grpc` grpc server.

## Issue reference

Closes #451 

## Checklist

* [x] Code compiles correctly
~* [ ] Created/updated tests~
~* [ ] Extended the documentation~

### DIFF
--- a/ext/dapr-ext-grpc/dapr/ext/grpc/app.py
+++ b/ext/dapr-ext-grpc/dapr/ext/grpc/app.py
@@ -44,10 +44,11 @@ class App:
         """
         self._servicer = _CallbackServicer()
         if not kwargs:
-            if max_grpc_message_length is None:
-                options = (
+            options = []
+            if max_grpc_message_length is not None:
+                options = [
                     ('grpc.max_send_message_length', max_grpc_message_length),
-                    ('grpc.max_receive_message_length', max_grpc_message_length),)
+                    ('grpc.max_receive_message_length', max_grpc_message_length)]
             self._server = grpc.server(  # type: ignore
                 futures.ThreadPoolExecutor(max_workers=10), options=options)
         else:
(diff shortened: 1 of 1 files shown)


ext/dapr-ext-grpc/dapr/ext/grpc/app.py
```diff
@@ -34,17 +34,25 @@ class App:
...
+            options = []
+            if max_grpc_message_length is not None:
+                options = [
+                    ('grpc.max_send_message_length', max_grpc_message_length),
+                    ('grpc.max_receive_message_length', max_grpc_message_length)]
+            self._server = grpc.server(  # type: ignore
+                futures.ThreadPoolExecutor(max_workers=10), options=options)
         else:
-            self._server = grpc.server(**kwargs)   # type: ignore
+            self._server = grpc.server(**kwargs)  # type: ignore
```
mukundansundar, line 55: @berndverst the issue here is this init function limits the options that can be set on the server when they also need to send message size... 

Can we get the max message size param from the kwargs and use that? That way users can use both kwargs and still be able to set other options?

mukundansundar (changes requested):

ext/dapr-ext-grpc/dapr/ext/grpc/app.py
mukundansundar: @berndverst  Nvm you already pointed that out in the issue.... The question is do we need this feature though? Why can't the users use the current way by setting options.... We can just document that is the way to use

mukundansundar (approved):
Lgtm

PR #460, opened 2022-09-20
Adds `/ok-to-test` option for workflow validation

berndverst:
Adds `/ok-to-test` to run example validation against latest `dapr/dapr`@master


mukundansundar (dismissed):

artursouza (approved):
I would recommend looking at how the e2e tests in Dapr post comments into the PR as a follow up.

PR #498, opened 2022-12-15
Set `ensure_ascii = False` in the default serializer

findmyway:
Signed-off-by: Jun Tian <find_my_way@foxmail.com>

# Description

With `ensure_ascii = False`, the Unicode is serialized as-is instead of `\u` escape sequence.

I believe this will make the serialized result more aligned with the default one in FastAPI.

https://github.com/encode/starlette/blob/d7558515020f4c0a96b199880c181a40678b18d5/starlette/responses.py#L197


berndverst (changes requested):
Fails the linter:
`./dapr/serializers/json.py:48:101: E501 line too long (105 > 100 characters)`

findmyway:
> Fails the linter: `./dapr/serializers/json.py:48:101: E501 line too long (105 > 100 characters)`

Updated. Try again?

berndverst:
Let's see if the test passes. If it all passes I'll merge now. Was busy fixing some other things - sorry for the delay.

findmyway:
Seems some test cases need to be updated accordingly.

berndverst:
> Seems some test cases need to be updated accordingly.

In that case, this might be a breaking change and I probably cannot accept your change.

findmyway:
> In that case, this might be a breaking change and I probably cannot accept your change. 

Yeah, I just realized that this is actually a breaking change.

This means we'd better consider merging this later in the next major version.

(Feel free to close this PR ;)

berndverst:
We already have a method / argument to provide your own serializer, maybe you can introduce a new optional boolean flag for `ensure_ascii` which defaults to `true`?

findmyway:
> We already have a method / argument to provide your own serializer, maybe you can introduce a new optional boolean flag for `ensure_ascii` which defaults to `true`?

I'm not sure which method you mean above. I add an `__init__` method with an argument of `ensure_ascii`. Is this what you mean?

berndverst:
> > We already have a method / argument to provide your own serializer, maybe you can introduce a new optional boolean flag for `ensure_ascii` which defaults to `true`?
> 
> 
> 
> I'm not sure which method you mean above. I add an `__init__` method with an argument of `ensure_ascii`. Is this what you mean?

__init__ of DaprHttpClient is what I meant.

That being said, in the Python we do most things via GRPC anyway.

findmyway:
> init of DaprHttpClient is what I meant.

1. The (de)serializer is used not only in `DaprHttpClient`, but also actors.
2. In the init of `DaprHttpClient`, the serializer is already instantiated.

> That being said, in the Python we do most things via GRPC anyway.

Well, in python, actors are still implemented via HTTP. And the initial goal of this PR is to make the state/response of actors more human-readable (when Unicode is included) while debugging.

berndverst:
code freeze for the next release will be in ~3 weeks in case you want to do something here

berndverst (approved):

berndverst:
Never mind, I think this should be good as is.

Issue #518, opened 2023-01-06
Setting deadLetterTopic in FastAPI/Flask


lucasalexsorensen:
## Describe the proposal
Currently, there seems to be missing a mechanism for setting the deadLetterTopic on subscriptions created through the FastAPI/Flask extensions.

Would it make sense to extend the respective function decorators such that the interface becomes something like the following?
@dapr_app.subscribe(pubsub="...", topic="...", **dead_letter_topic="..."**)

lucasalexsorensen:
Created a draft pull request here for this:
https://github.com/dapr/python-sdk/pull/519

PR #519, opened 2023-01-06
dead_letter_topic for @subscribe (fastAPI/flask)

lucasalexsorensen:
# Description

Added an additional argument 'dead_letter_topic' to the @subscribe decorators for the FastAPI and Flask extensions.

## Issue reference

Please reference the issue this PR will close: #518

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/examples/pubsub-simple/subscriber.py
+++ b/examples/pubsub-simple/subscriber.py
@@ -60,4 +60,4 @@ def mytopic_wildcard(event: v1.Event) -> TopicEventResponse:
     return TopicEventResponse('success')
 
 
-app.run(50071)
+app.run(50051)
--- a/examples/pubsub-simple/subscriber.py
+++ b/examples/pubsub-simple/subscriber.py
@@ -23,7 +23,7 @@ app = App()
 should_retry = True  # To control whether dapr should retry sending a message
 
 
-@app.subscribe(pubsub_name='pubsub', topic='TOPIC_A', dead_letter_topic='TOPIC_A_DEAD')
+@app.subscribe(pubsub_name='pubsub', topic='TOPIC_A')
 def mytopic(event: v1.Event) -> TopicEventResponse:
     global should_retry
     data = json.loads(event.Data())
@@ -35,7 +35,13 @@ def mytopic(event: v1.Event) -> TopicEventResponse:
         return TopicEventResponse('retry')
     return TopicEventResponse('success')
 
-@app.subscribe(pubsub_name='pubsub', topic='TOPIC_A_DEAD')
+
+@app.subscribe(pubsub_name='pubsub', topic='TOPIC_D', dead_letter_topic='TOPIC_D_DEAD')
+def fail_and_send_to_dead_topic(event: v1.Event) -> TopicEventResponse:
+    return TopicEventResponse('retry')
+
+
+@app.subscribe(pubsub_name='pubsub', topic='TOPIC_D_DEAD')
 def mytopic_dead(event: v1.Event) -> TopicEventResponse:
     data = json.loads(event.Data())
     print(f'Dead-Letter Subscriber received: id={data["id"]}, message="{data["message"]}", '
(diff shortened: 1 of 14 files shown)


berndverst:
PR looks good. Can you also run the linter please `tox -e flake8`. There are some linter problems - hence the "build" failures.

Could you also increase the scope and implement this for the `gRPC` case (`dapr.ext.grpc`)?
- in class `_CallBackServicer` update `register_topic` to take an Optional `dead_letter_topic`.
https://github.com/dapr/python-sdk/blob/edff0ff51b864e24048d784fff16153fa8f2a313/ext/dapr-ext-grpc/dapr/ext/grpc/_servicier.py#L74-L81
Then in line 100 just provide the `dead_letter_topic` string parameter if the value wasn't `None`.
https://github.com/dapr/python-sdk/blob/edff0ff51b864e24048d784fff16153fa8f2a313/ext/dapr-ext-grpc/dapr/ext/grpc/_servicier.py#L100-L105

And then add the optional param to the decorator here:
https://github.com/dapr/python-sdk/blob/edff0ff51b864e24048d784fff16153fa8f2a313/ext/dapr-ext-grpc/dapr/ext/grpc/app.py#L118-L140

Finally, it would be cool if you could update the `pubsub-simple` example which uses the gRPC client. All you'd have to do is provide the additional `dead_letter_topic` parameter to the annotation. Feel free to make up a topic name .. something that is reasonable for this example.

https://github.com/dapr/python-sdk/blob/edff0ff51b864e24048d784fff16153fa8f2a313/examples/pubsub-simple/subscriber.py#L48-L54

ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py
```diff
@@ -83,7 +85,8 @@ def decorator(func):
                 "pubsubname": pubsub,
                 "topic": topic,
                 "route": event_handler_route,
-                "metadata": metadata
+                "metadata": metadata,
+                **({"deadLetterTopic": dead_letter_topic} if dead_letter_topic else {})
```
berndverst, line 89: I didn't realize this syntax can be used in this context. TIL -- very useful!

I'd probably have written another.. `if dead_letter_topic append...`

lucasalexsorensen:
I've extended it for gRPC, updated the pubsub example, updated a few tests and fixed the linter errors (also fixed the DCO signoff thing)

Let me know if anything else needs changing :)

ext/dapr-ext-fastapi/dapr/ext/fastapi/app.py
berndverst: Works for me!

examples/pubsub-simple/subscriber.py
```diff
@@ -54,4 +60,4 @@ def mytopic_wildcard(event: v1.Event) -> TopicEventResponse:
     return TopicEventResponse('success')
 
 
-app.run(50051)
+app.run(50071)
```
berndverst, line 63:
```suggestion
app.run(50051)
```

It needs to be port 50051 because the corresponding test README file has an instruction to start Dapr with this app port.

berndverst (changes requested):
Apparently providing a deadletter fails the RETRY test -- so I updated the example to distinguish retry test from deadletter test.

Normally I have permission to just commit to the PR author's fork - but you didn't give me this permission.

I prepared a patch file you can just download and apply.
https://github.com/berndverst/python-sdk/commit/da0fae31b9e3f7027e320e314f3aed297a396ce1.patch

```
wget https://github.com/berndverst/python-sdk/commit/da0fae31b9e3f7027e320e314f3aed297a396ce1.patch
git apply da0fae31b9e3f7027e320e314f3aed297a396ce1.patch
```

Don't forget to create a commit with the changed files in `examples/pubsub-simple` and commit that.

Once you applied the patch this PR should be good to go.

berndverst (approved):

lucasalexsorensen:
Awesome, thanks 😃

PR #521, opened 2023-01-07
Fix return type of resource enter protocol in DaprGrpcClient. Fixes #520

### LINKED ISSUE #520: The DaprClient resource manager returns wrong type
## Expected Behavior

Using `with` statement with `DaprClient` should return type of `DaprClient` on enter. To make this happen the `__enter__` method of `DaprGrpcClient` should return type `Self` and not `DaprGrpcClient` so that sub-classes return the correct type. E.g `DaprClient` would return type of `DaprClient` https://github.com/dapr/python-sdk/blob/edff0ff51b864e24048d784fff16153fa8f2a313/dapr/clients/grpc/client.py#L150

The `Self` type can be used from Python 3.11. For versions less than 3.11 it can be imported from `typing_extensions` e.g:

```py
from typing_extensions import Self
...

class DaprGrpcClient:
    ....
    def __enter__(self) -> Self:
        return self
```

## Actual Behavior

Using `with` statement with `DaprClient` returns type of `DaprGrpcClient`, not `DaprClient`. This means that methods like e.g `invoke_method_async` is not available for static type checkers such as Pyright.

Note that this is a static type checker problem, not a runtime problem. Code runs just fine, but editor shows a problem.

## Steps to Reproduce the Problem

Writing code such as `with DaprClient() as d:` in editors such as vscode using standard Python extension with Pylance (pyright) will show error.

Screenshot 2023-01-07 at 15 04 13

## Release Note

RELEASE NOTE:

dbrattli:
# Description

Fixes return type of `DaprGrpcClient` so that sub-classes i.e `DaprClient` gets the right type back when doing `with DaprClient as d`.

Note that mypy does not currently handle the `Self` type https://github.com/python/mypy/issues/14089 so we need a new version of mypy for this to work. Current version of mypy gives: `Variable "typing_extensions.Self" is not valid as a type  [valid-type]`. To make it work I've added `# type: ignore` on the line which seems to work with both current version of mypy and sub-classes gets the right type with type checkers such as pyright.

## Issue reference

Please reference the issue this PR will close: #520 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests (not needed)
* [ ] Extended the documentation (not needed)

### DIFF
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -21,6 +21,7 @@ from urllib.parse import urlencode
 from warnings import warn
 
 from typing import Dict, Optional, Union, Sequence, List
+from typing_extensions import Self
 
 from google.protobuf.message import Message as GrpcMessage
 from google.protobuf.empty_pb2 import Empty as GrpcEmpty
@@ -147,7 +148,7 @@ class DaprGrpcClient:
     def __del__(self):
         self.close()
 
-    def __enter__(self) -> 'DaprGrpcClient':
+    def __enter__(self) -> Self:  # type: ignore
         return self
 
     def __exit__(self, exc_type, exc_value, traceback) -> None:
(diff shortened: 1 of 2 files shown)


berndverst (approved):
Looks good to me. Thanks for the contribution.

berndverst (approved):

PR #475, opened 2022-10-06
Add PEP 561 py.typed markers to all packages to allow consumers to check dapr types with e.g. mypy

Trolldemorted:
# Description

This PR adds PEP 561 py.typed markers to all packages to indicate that downstream static code analysis tools may use the type annotations for static code analysis.

Given this (correct) code:
```python
from dapr.clients import DaprClient
with DaprClient() as d:
    d.publish_event(
        pubsub_name="pubsub_name",
        topic_name="topic_name",
        data="",
        data_content_type="application/json",
    )
```
`python3 -m mypy main.py` succeeds (`Success: no issues found in 1 source file`) as expected.

Given this (wrong) code:
```python
from dapr.clients import DaprClient
with DaprClient() as d:
    d.publish_event(
        pubsub_name=5,
        topic_name="topic_name",
        data="",
        data_content_type="application/json",
    )
```
`python3 -m mypy main.py` fails as expected and yields the expected error message:
```
main.py:4: error: Argument "pubsub_name" to "publish_event" of "DaprGrpcClient" has incompatible type "int"; expected "str"
Found 1 error in 1 file (checked 1 source file)
```

## Issue reference
Fixes #456

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

I haven taken the liberty to not touch tests or documentation, because the change is very simple.

PR #532, opened 2023-01-20
more mypy markers

Trolldemorted:
# Description

Added py.typed markers for the ext modules.

## Issue reference

#475 did not cover all modules, so here we go again :)


Trolldemorted:
looooks like this doesn't work as smoothly, sorry :(

PR #553, opened 2023-04-15
Async support for Python SDK (`dapr.aio.clients.DaprClient`)

### LINKED ISSUE #90: Add support for async gRPC in Dapr SDK

berndverst:
# Description

This adds an asynchronous client `DaprClient` under new package `dapr.aio.clients`, which under the hood uses `dapr.aio.clients.grpc.GrpcDaprClientAsync`.

All client methods are asynchronous.

Example
```python
from dapr.aio.clients import DaprClient
async with DaprClient() as d:
  result = await d.get_metadata()
  print(result.application_id)
```

Tests have been added to `tests/clients/test_dapr_async_grpc_client.py`

Fixes #90


tests/clients/test_dapr_async_grpc_client.py
```diff
@@ -0,0 +1,759 @@
+# -*- coding: utf-8 -*-
```
berndverst, line 1: **Reviewers:** This file copies the existing synchronous tests and rewrites them as async using the async client.

No new tests have been introduced here.

dapr/aio/clients/grpc/client.py
```diff
@@ -0,0 +1,1227 @@
+# -*- coding: utf-8 -*-
```
berndverst, line 1: Reviewers: This file is a copy from dapr.clients.grpc client.py but uses the asynchronous grpc package and imports other async helpers. All methods have been updated for asynchronous usage.

No new functionality from the synchronous has been added.

dapr/aio/clients/grpc/_asynchelpers.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
```
berndverst, line 1: Reviewers: This file contains two classes from dapr.clients.grpc._helpers adjusted for asynchronous grpc use. These classes are almost identical to their synchronous version.

mukundansundar (approved):

Issue #542, opened 2023-03-13
Workflow implementation in SDK


RyanLettieri:
## Describe the proposal

Add python SDK support for dapr workflow. This will include both the authoring and management functionalities. 

Related dapr/dapr proposal: https://github.com/dapr/dapr/issues/4576

berndverst:
@RyanLettieri make sure you add the implementation in both `dapr/clients/grpc/client.py` and `dapr/aio/clients/grpc/client.py` (the latter being the async client)

PR #554, opened 2023-04-18
Workflow management python SDK

RyanLettieri:
# Description

This PR covers the implementation of the management portion for workflow.

Note that this PR will fail until the following PRs are submitted:
https://github.com/dapr/dapr/pull/6163
https://github.com/dapr/components-contrib/pull/2729

## Issue reference

Please reference the issue this PR will close: #542

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


examples/workflow/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+- [Dapr CLI and initialized environment](https://docs.dapr.io/getting-started)
+- [Install Python 3.7+](https://www.python.org/downloads/)
+
+## Install Dapr python-SDK
+
+<!-- Our CI/CD pipeline automatically installs the correct version, so we can skip this step in the automation -->
+
+```bash
+pip3 install dapr dapr-ext-grpc
```
berndverst, line 28:
```suggestion
pip3 install dapr
```

You don't need to just a grpc server, so you don't need that dependency

berndverst:
```
dapr/clients/grpc/_response.py:963: error: Too many arguments for "__init__" of "object"  [call-arg]
dapr/clients/grpc/_response.py:989: error: Too many arguments for "__init__" of "object"  [call-arg]
```

dapr/clients/grpc/_response.py
```diff
@@ -933,6 +933,64 @@ def extended_metadata(self) -> Dict[str, str]:
...
+        headers: MetadataTuple = (),
+    ):
+        """Initializes a StartWorkflowResponse.
+
+        Args:
+            instance_id (str): the instance ID assocated with this response.
+            client (DaprClient): a reference to the dapr client used for the GetWorkflow request.
+            headers (Tuple, optional): the headers from Dapr gRPC response.
+        """
+        super().__init__(headers)
```
berndverst, line 989: You cannot call super because your class doesn't extend anything. Did you mean to extend DaprResponse?

dapr/clients/grpc/_response.py
```diff
@@ -933,6 +933,64 @@ def extended_metadata(self) -> Dict[str, str]:
...
+            client (DaprClient): a reference to the dapr client used for the GetWorkflow request.
+            instance_id (str): the instance ID assocated with this response.
+            workflow_name (str): the name of the workflow that was started.
+            created_at (str): the time at which the workflow started executing.
+            last_updated_at (str): the time at which the workflow was last updated.
+            runtime_status (str): the current runtime status of the workflow.
+            properties (Dict[str, str]): properties sent as a reponse by the workflow.
+            headers (Tuple, optional): the headers from Dapr gRPC response.
+        """
+        super().__init__(headers)
```
berndverst, line 963: You cannot call super because you are not extending a parent class here. Did you mean to use DaprResponse?

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component)
+        # Actual get workflow invocation
+        req = api_v1.GetWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component)
+        response, call = self._stub.GetWorkflowAlpha1.with_call(req)
+
+        # RRL TODO: Fix this
+        return api_v1.GetWorkflowResponse(instance_id=instance_id,
```
berndverst, line 1157: You should return the actual GetWorkflowResponse class you created.

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+        # Actual start workflow invocation
+        req = api_v1.StartWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component,
+            workflow_name=workflow_name,
+            options=workflow_options,
+            input=input)
+        response, call = self._stub.StartWorkflowAlpha1.with_call(req)
+
+        return api_v1.StartWorkflowResponse(instance_id=response.instanceID)
```
berndverst, line 1127: This should return the StartWorkflowResponse from your response helpers - but you also need to import that.

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component)
+        # Actual terminate workflow invocation
+        req = api_v1.TerminateWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component)
+        _, call = self._stub.TerminateWorkflowAlpha1.with_call(req)
+
+        return DaprResponse(
```
berndverst, line 1190: This is a super generic response - how does someone know whether the termination request was received properly?

If there something in the protos that provides a better response?

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+            headers=call.initial_metadata())
+
+    # RRL TODO: Clean up return obkect comments
```
berndverst, line 1193: fix comment - also what does this mean?

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+        # Warnings and input validation
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component)
+        # Actual terminate workflow invocation
+        req = api_v1.TerminateWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component)
+        _, call = self._stub.TerminateWorkflowAlpha1.with_call(req)
```
berndverst, line 1218: Why does this also call TerminateWorkflow? I thought Raise Event is something else?

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+
+        return DaprResponse(
```
berndverst, line 1220: Super generic response - is there something better?

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component)
+        # Actual pause workflow invocation
+        req = api_v1.PauseWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component)
+        _, call = self._stub.PauseWorkflowAlpha1.with_call(req)
+
+        return DaprResponse(
```
berndverst, line 1250: Is there a less generic response indicating success?

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component)
+        # Actual resume workflow invocation
+        req = api_v1.ResumeWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component)
+        _, call = self._stub.ResumeWorkflowAlpha1.with_call(req)
+
+        return DaprResponse(
```
berndverst, line 1279: Ditto

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1088,226 @@ def unlock(
...
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component)
+        # Actual purge workflow invocation
+        req = api_v1.PurgeWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component)
+        _, call = self._stub.PurgeWorkflowAlpha1.with_call(req)
+
+        return DaprResponse(
```
berndverst, line 1308: Ditto

examples/workflow/workflow.py
```diff
@@ -0,0 +1,29 @@
...
+
+with DaprClient() as d:
+    workflowName = 'OrderProcessingWorkflow'
+    workflowComponent = 'dapr'
+    instanceId = 'testInstance'
+
+    inventoryItem = "Computer"
+    ammount = "1"
+
+
```
berndverst, line 16:
```suggestion
```

tests/clients/test_dapr_grpc_client.py
```diff
@@ -681,6 +681,53 @@ def test_unlock_input_validation(self):
...
+    #
+    # Tests for workflow
+    #
+
+    def test_workflow(self):
+        dapr = DaprGrpcClient(f'localhost:{self.server_port}')
+        # Sane parameters
+        workflow_name = 'testWorkflow'
+        instance_id = str(uuid.uuid4())
+        workflow_component = "dapr"
```
berndverst, line 693: If the workflow component were something else (not `dapr`) this test would fail. This code here relies on the sidecar running - and that's not what we want for our unit tests. Only for integration tests!

I would like this to be a mocked test against the Mock gRPC server instead.

You need to stub out methods and provide expected responses here:
`tests/clients/fake_dapr_server.py`

The example in the examples folder can act as your integration test running against the real workflow engine. Those tests are always run with a real sidecar.

berndverst (changes requested):
Left a bunch of change requests.

At a high level:

- Do not use super generic response classes if they can be avoided.
- Do not use the response classes from the protos themselves, instead create your own more Pythonic response classes (see `_response.py`)
- In unit tests, do not make calls against real sidecar - the sidecar will not be running. Instead make sure to add methods to the fake dapr sidecar in `fake_dapr_server.py` (gRPC) and `fake_http_server.py` (HTTP) if applicable.

dapr/clients/grpc/_response.py
berndverst: This is probably the root cause of the Type check error by the way (`tox -e type`)

berndverst (changes requested):
Running your workflow test I get:

```
		== APP == Traceback (most recent call last):
		== APP ==   File "/Users/bverst/src/dapr/python-sdk/examples/workflow/workflow.py", line 20, in <module>
		== APP ==     d.start_workflow(instance_id=instanceId, workflow_component=workflowComponent,
		== APP == TypeError: DaprGrpcClient.start_workflow() missing 1 required positional argument: 'workflow_options'
		ℹ️
		terminated signal received: shutting down
		✅  Exited Dapr successfully

	ERROR expected lines not found:
		== APP == State store has successfully saved value_1 with key_1 as key
		== APP == Cannot save due to bad etag. ErrorCode=StatusCode.ABORTED
		== APP == State store has successfully saved value_2 with key_2 as key
		== APP == State store has successfully saved value_3 with key_3 as key
		== APP == Cannot save bulk due to bad etags. ErrorCode=StatusCode.ABORTED
		== APP == Got value=b'value_1' eTag=1
		== APP == Got items with etags: [(b'value_1_updated', '2'), (b'value_2', '2')]
		== APP == Got value after delete: b''
		```

**Comment by berndverst** [comment]
It seems you should have made `workflow_options` an `Optional` type and set the default value to `None`.

File: dapr/aio/clients/grpc/client.py [hunk 15]
```diff
@@ -1103,6 +1105,237 @@ async def unlock(
...
+        """Starts a workflow.
+
+            Args:
+                instance_id (str): the name of the workflow instance,
+                                    e.g. `order_processing_workflow-103784`.
+                workflow_component (str): the name of the workflow component
+                                    that will run the workflow. e.g. `dapr`.
+                workflow_name (str): the name of the workflow that will be executed.
+                input (bytes): the input that the workflow will receive.
+                workflow_options (dict): the key-value options that the workflow will receive.
```
berndverst, line 1124:
```suggestion
                workflow_options (dict, optional): the key-value options that the workflow will receive.
```

dapr/clients/grpc/_response.py
```diff
@@ -933,6 +933,52 @@ def extended_metadata(self) -> Dict[str, str]:
         return self._extended_metadata
 
 
+# RRL TODO: Add properties and fix init
```
berndverst, line 936: Fix or remove in this PR

tests/clients/fake_dapr_server.py
```diff
@@ -259,6 +268,66 @@ def UnlockAlpha1(self, request: UnlockRequest, context):
...
 
+    def StartWorkflowAlpha1(self, request: StartWorkflowRequest, context):
+        instance_id = (request.instance_id)
+
+        if instance_id not in self.running_workflows:
+            self.running_workflows[instance_id] = "RUNNING"
+            return StartWorkflowResponse(instance_id=instance_id)
+        else:
+            # workflow already running
+            return False  # RRL TODO: Determine how to return an error or value of FALSE here
```
berndverst, line 279: You can make it so that a particular parameter passed with a value ending in `_fail` or whatever you want triggers the failure case.

tests/clients/fake_dapr_server.py
```diff
@@ -259,6 +268,66 @@ def UnlockAlpha1(self, request: UnlockRequest, context):
...
+
+    def GetWorkflowAlpha1(self, request: GetWorkflowRequest, context):
+        instance_id = (request.instance_id)
+
+        if instance_id in self.running_workflows:
+            return GetWorkflowResponse(instance_id=instance_id,
+                                       runtime_status=self.running_workflows[instance_id])
+        else:
+            # workflow non-existent
+            return False  # RRL TODO: Determine how to return an error or value of FALSE here
```
berndverst, line 289: You can make it so that a particular parameter passed with a value ending in `_fail` or whatever you want triggers the failure case.

tests/clients/fake_dapr_server.py
```diff
@@ -259,6 +268,66 @@ def UnlockAlpha1(self, request: UnlockRequest, context):
...
+
+    def PauseWorkflowAlpha1(self, request: PauseWorkflowRequest, context):
+        instance_id = (request.instance_id)
+
+        if instance_id in self.running_workflows:
+            self.running_workflows[instance_id] = "SUSPENDED"
+            return empty_pb2.Empty()
+        else:
+            # workflow non-existent
+            return empty_pb2.Empty()  # RRL TODO: Determine how to return an error or FALSE here
```
berndverst, line 299: Ditto: Use something from the pause workflow request params to trigger the failure case here.

tests/clients/fake_dapr_server.py
```diff
@@ -259,6 +268,66 @@ def UnlockAlpha1(self, request: UnlockRequest, context):
...
+
+    def ResumeWorkflowAlpha1(self, request: ResumeWorkflowRequest, context):
+        instance_id = (request.instance_id)
+
+        if instance_id in self.running_workflows:
+            self.running_workflows[instance_id] = "RUNNING"
+            return empty_pb2.Empty()
+        else:
+            # workflow non-existent
+            return empty_pb2.Empty()  # RRL TODO: Determine how to return an error or FALSE here
```
berndverst, line 309: Ditto: Use custom metadata / values to trigger a failure case.

berndverst (changes requested):

berndverst (changes requested):
A few more unit test changes required.

Also we need to figure out what to do about the examples integration test.

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1093,263 @@ def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=call.initial_metadata())
 
+    def start_workflow(
```
tmacam, line 1096: Aside the exception being caught in the catch blocks, are there major differences between the AIO and non-AIO versions? Could this be in a mixin class?

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1093,263 @@ def unlock(
...
+            Returns:
+                :class:`DaprResponse` gRPC metadata returned from callee
+        """
+        # Warnings and input validation
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component,
+                               event_name=event_name)
+        # Actual terminate workflow invocation
```
tmacam, line 1234:
```suggestion
        # Actual raise_event workflow invocation
```

dapr/clients/grpc/client.py
tmacam: NVM, we tag the async functions with `async`. A mixin would not work here :(

dapr/clients/grpc/_response.py
```diff
@@ -933,6 +933,51 @@ def extended_metadata(self) -> Dict[str, str]:
...
 
+class GetWorkflowResponse():
+    '''The response of get_workflow operation.'''
+
+    def __init__(
+        self,
+        instance_id: str,
+        workflow_name: str,
+        created_at: str,
+        last_updated_at: str,
```
cgillum, line 944: Both `created_at` and `last_updated_at` should be `datetime`, not `str`.

dapr/clients/grpc/_response.py
```diff
@@ -933,6 +933,51 @@ def extended_metadata(self) -> Dict[str, str]:
...
+        runtime_status: str,
```
cgillum, line 945: This should be `WorkflowRuntimeStatus`, not `str`

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1094,258 @@ def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=call.initial_metadata())
 
+    def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: bytes,
```
cgillum, line 1101: Same comment about `bytes` vs `Any`.

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1094,258 @@ def unlock(
...
+        # Actual start workflow invocation
+        req = api_v1.StartWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component,
+            workflow_name=workflow_name,
+            options=workflow_options,
+            input=encoded_data)
+
+        try:
+            response, call = self._stub.StartWorkflowAlpha1.with_call(req)
```
cgillum, line 1133: Same question about "call" here and in all the other methods.

examples/workflow/README.md
```diff
@@ -0,0 +1,64 @@
+# Example - workflow
+
+This example demonstrates the workflow APIs in Dapr.
+It demonstrates the following APIs:
+- **start_workflow**: Start an instance of a workflow
+- **get_workflow**: Get information on a single workflow
+- **terminate_workflow**: Terminate or stop a particular instance of a workflow
+- **raise_event**: Raise an event on a workflow
+- **pause_workflow**: Pauses or suspends a workflow instance that can later be resumed
```
cgillum, line 9:
```suggestion
- **pause_workflow**: Pauses a workflow instance
```

examples/workflow/README.md
```diff
@@ -0,0 +1,64 @@
...
+- **resume_workflow**: Resumes a paused workflow instance
+- **purge_workflow**: Removes all metadata related to a specific workflow instance
```
cgillum, line 11:
```suggestion
- **purge_workflow**: Removes all metadata related to a specific workflow instance from the state store
```

examples/workflow/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+It creates a client using `DaprClient` and calls all the state API methods available as example.
+It uses the default configuration from Dapr init in [self-hosted mode](https://github.com/dapr/cli#install-dapr-on-your-local-machine-self-hosted). 
+
+> **Note:** Make sure to use the latest proto bindings
```
cgillum, line 16: What does this mean?

examples/workflow/workflow.py
```diff
@@ -0,0 +1,80 @@
...
+    nonExistentIDError = "No such instance exists"
+    # Wait for sidecar to be up within 5 seconds.
+    d.wait(5)
+
+    sleep(5)
+
+    # Start the workflow
+    start_resp = d.start_workflow(instance_id=instanceId, workflow_component=workflowComponent,
+                     workflow_name=workflowName, input=item2, workflow_options=workflowOptions)
+    print(f"Attempting to start {workflowName}")
```
cgillum, line 34: This `print` should go before the call to `start_workflow`.

examples/workflow/workflow.py
```diff
@@ -0,0 +1,80 @@
...
+    print(f"start_resp {start_resp.instance_id}")
+    getResponse = d.get_workflow(instance_id=instanceId, workflow_component=workflowComponent)
+    print(f"Get response from {workflowName} after start call: {getResponse.runtime_status}")
+
+    # Raise event
+    d.raise_workflow_event(instance_id=instanceId, workflow_component=workflowComponent,
+                  event_name=eventName, event_data=eventData)
+    
+    # Sleep so that the workflow can finish 
+    sleep(5)
```
cgillum, line 77: Consider adding a `wait_for_workflow_completion` method, like we have [here in the .NET SDK](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Client/DaprClient.cs#L1051). This is also similar to the `d.wait(5)` method that you're using above to wait for the dapr sidecar to start.

tests/clients/fake_dapr_server.py
```diff
@@ -259,6 +270,76 @@ def UnlockAlpha1(self, request: UnlockRequest, context):
         else:
             return UnlockResponse(status=UnlockResponse.Status.LOCK_BELONGS_TO_OTHERS)
 
+    def StartWorkflowAlpha1(self, request: StartWorkflowRequest, context):
+        instance_id = (request.instance_id)
```
cgillum, line 274: Why do you have parenthesis here?

tests/clients/test_dapr_grpc_client.py
```diff
@@ -681,6 +682,58 @@ def test_unlock_input_validation(self):
...
+        # Sane parameters
+        workflow_name = 'testWorkflow'
+        event_name = 'eventName'
+        instance_id = str(uuid.uuid4())
+        workflow_component = "dapr"
+        input = "paperclips"
+        event_data = "cars"
+        # Start the workflow
+        start_response = dapr.start_workflow(instance_id, workflow_name, workflow_component,
+                                             input.encode('utf-8'), None)
```
cgillum, line 700: Related to my other comment, I think we want to allow users to just pass `input` here.

tests/clients/test_dapr_grpc_client.py
```diff
@@ -681,6 +682,58 @@ def test_unlock_input_validation(self):
...
+        self.assertEqual(WorkflowRuntimeStatus.SUSPENDED, get_response.runtime_status)
+
+        # Resume the workflow
+        dapr.resume_workflow(instance_id, workflow_component)
+
+        # Get info on the workflow to check that it is resumed
+        get_response = dapr.get_workflow(instance_id, workflow_component)
+        self.assertEqual(WorkflowRuntimeStatus.RUNNING, get_response.runtime_status)
+
+        # Raise an event on the workflow. TODO: Figure out how to check/verify this
```
cgillum, line 721: Regarding this TODO, consider putting some data in the workflow response properties collection to let you know that the raise_event call succeeded.

cgillum (changes requested):

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1108,260 @@ async def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=await call.initial_metadata())
 
+    async def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: any,
+            instance_id: Optional[str] = "",
```
cgillum, line 1116: The default value should be `None` when using `Optional[str]`.

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1108,260 @@ async def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=await call.initial_metadata())
 
+    async def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: Optional[bytes],
```
cgillum, line 1115: `bytes` is not a user-friendly input type. Most users would prefer `Any` and rely on use to do JSON serialization since that's what the workflow authoring SDKs generally expect. I think we can support both by making this `Union[Any, bytes, None] = None`.
```suggestion
            input: Union[Any, bytes, None] = None,
```

Then, down below, only serialize to JSON if the input type is something other than `bytes` - i.e.

```python
if isinstance(input, bytes):
    encoded_data = bytes
else:
    encoded_data = json.dumps(input).encode("utf-8") if input is not None else None
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1108,260 @@ async def unlock(
...
+        # Actual start workflow invocation
+        req = api_v1.StartWorkflowRequest(
+            instance_id=instance_id,
+            workflow_component=workflow_component,
+            workflow_name=workflow_name,
+            options=workflow_options,
+            input=encoded_data)
+
+        try:
+            response, _ = self._stub.StartWorkflowAlpha1.with_call(req)
```
cgillum, line 1149: Why are we using `with_call`? Why can't we just do `self._stub.StartWorkflowAlpha1(req)`? I think the point of my previous comment on this was missed.

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1108,260 @@ async def unlock(
...
+                headers=call.initial_metadata())
+        except grpc.aio.AioRpcError as err:
+            raise DaprInternalError(err.details())
+
+    async def raise_workflow_event(
+            self,
+            instance_id: str,
+            workflow_component: str,
+            event_name: str,
+            event_data: Optional[bytes]) -> DaprResponse:
```
cgillum, line 1228:
```suggestion
            event_data: Optional[bytes] = None) -> DaprResponse:
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1108,260 @@ async def unlock(
...
+            Returns:
+                :class:`DaprResponse` gRPC metadata returned from callee
+        """
+        # Warnings and input validation
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component,
+                               event_name=event_name)
+        encoded_data = json.dumps(event_data).encode("UTF-8")
```
cgillum, line 1249: I think you need a `None` check here.

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1094,259 @@ def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=call.initial_metadata())
 
+    def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: Optional[bytes],
```
cgillum, line 1101: Same feedback in this file as in aoi/clients/grpc/client.py.

dapr/clients/grpc/_response.py
```diff
@@ -933,6 +933,52 @@ def extended_metadata(self) -> Dict[str, str]:
...
+    def __init__(
+        self,
+        instance_id: str,
+    ):
+        """Initializes a StartWorkflowResponse.
+
+        Args:
+            instance_id (str): the instance ID assocated with this response.
+        """
+        self.instance_id = instance_id
```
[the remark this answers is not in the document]
cgillum: [@dataclass](https://docs.python.org/3/library/dataclasses.html) is another option, though I see that we already used `NamedTuple` in this codebase.

examples/demo_workflow/app.py
```diff
@@ -32,32 +43,61 @@ def hello_act(ctx: WorkflowActivityContext, input):
...
-    assert status.runtime_status.name == "COMPLETED"
-    workflowRuntime.shutdown()
+    with DaprClient() as d:
+        workflowRuntime = WorkflowRuntime()
+        workflowRuntime.register_workflow(hello_world_wf)
+        workflowRuntime.register_activity(hello_act)
+        workflowRuntime.start()
+
+        host = settings.DAPR_RUNTIME_HOST
+        if host is None:
```
cgillum, line 53: `host` will never be `None`

examples/demo_workflow/app.py
```diff
@@ -32,32 +43,61 @@ def hello_act(ctx: WorkflowActivityContext, input):
...
+            host = "localhost"
+        port = settings.DAPR_GRPC_PORT
+        if port is None:
+            port = "4001"
```
cgillum, line 57: `port` will never be `None`.

tests/clients/fake_dapr_server.py
```diff
@@ -28,6 +38,8 @@ def __init__(self):
         self.store = {}
         self.shutdown_received = False
         self.locks_to_owner = {}  # (store_name, resource_id) -> lock_owner
+        self.workflow_status = {}
+        self.workflow_options: Dict[Dict[str]] = dict()
```
cgillum, line 42: Is this a typo? What is a `Dict[Dict[str]]`?

dapr/aio/clients/grpc/client.py
RyanLettieri: Our API inside dapr/dapr checks against an empty string, shouldn't we leave it as "" rather than None?
Furthermore, the linter isn't the biggest fan of the change and gives the following error:
"Incompatible default for argument "instance_id" (default has type "None", argument has type "str")"

cgillum (approved):
LGTM

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=await call.initial_metadata())
 
+    async def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: Union[Any, bytes, None] = None,
+            instance_id: Optional[str] = None,
+            workflow_options: Optional[Dict[str, str]] = dict()) -> StartWorkflowResponse:
```
berndverst, line 1119: If you are assigning a default value of `dict()` anyway, then you don't need this to be `Optional` - only use `Optional` when you want to set it to `None` by default or you want to explicitly support `None`.

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
...
+        """
+        # Warnings and input validation
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component,
+                               workflow_name=workflow_name)
+
+        if instance_id is None:
+            instance_id = str(uuid.uuid4())
```
berndverst, line 1143: This really should have been handled in the runtime. Can you make it so that you can send an empty string and the runtime generates a UUID in that case (in the future)? @cgillum

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
...
+
+        if isinstance(input, bytes):
+            encoded_data = input
+        else:
+            encoded_data = json.dumps(input).encode("utf-8") if input is not None else bytes([])
```
berndverst, line 1148: This will cause completely arbitrary results converting who knows what kind of object to bytes this way.

I strongly recommend removing this together with the `Any` type and letting a user handle the bytes conversion themselves (so that they know what conversion they used)

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
...
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component,
+                               event_name=event_name)
+        if isinstance(input, bytes):
+            encoded_data = event_data
+        else:
+            if event_data is not None:
+                encoded_data = json.dumps(event_data).encode("utf-8")
```
berndverst, line 1267: Same here - this can cause unpredictable results. Remove `Any` and and this case. Only support `None` and `bytes` as types. The user needs to explictly perform the conversion in their own code.

berndverst (changes requested):

berndverst (commented):
The Python SDK always uses `Union[bytes, str]` for any value which is `bytes` in the proto.

We should maintain this.

So if the data is `str` you probably need to convert it to `bytes` first. The `bytes` and `None` cases are otherwise handled correctly.

It is better to let a user manually define how they wish to convert any other object to bytes or string themselves, rather than doing this for the user. This will reduce a lot of support burden as users will be confused how another object type was encoded and then be unable to decode it.

dapr/aio/clients/grpc/client.py
cgillum: We handled this in the HTTP layer but not the gRPC layer. I've opened an issue to track fixing this in the gRPC layer: https://github.com/dapr/dapr/issues/6425.

berndverst (commented):
Here is a better idea:

If you replace `Any` with `Dict` then you still have convenience for the user, but you don't support completely arbitrary type conversions (which is a bad idea).

Please update the signature to `Optional[Union[Dict, bytes, str]]`

I believe other than replacing Any for Dict, and changing this signature (and doc strings) you might not need to make any other code changes here.

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
...
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component,
+                               workflow_name=workflow_name)
+
+        if instance_id is None:
+            instance_id = str(uuid.uuid4())
+
+        if isinstance(input, bytes):
```
berndverst, line 1145:
```suggestion
        if send_raw_bytes and isinstance(input, bytes):
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=await call.initial_metadata())
 
+    async def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: Union[Any, bytes, None] = None,
+            instance_id: Optional[str] = None,
```
berndverst, line 1118: also add parameter `send_raw_bytes: bool = False`

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
...
+                headers=call.initial_metadata())
+        except grpc.aio.AioRpcError as err:
+            raise DaprInternalError(err.details())
+
+    async def raise_workflow_event(
+            self,
+            instance_id: str,
+            workflow_component: str,
+            event_name: str,
+            event_data: Union[Any, bytes, None] = None) -> DaprResponse:
```
berndverst:
```suggestion
            event_data: Optional[Union[Any, bytes]] = None) -> DaprResponse:
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=await call.initial_metadata())
 
+    async def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: Union[Any, bytes, None] = None,
```
berndverst, line 1117:
```suggestion
            input: Optional[Union[Any, bytes]] = None,
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
...
+            Returns:
+                :class:`DaprResponse` gRPC metadata returned from callee
+        """
+        # Warnings and input validation
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component,
+                               event_name=event_name)
+        if isinstance(input, bytes):
```
berndverst:
```suggestion
        if `send_raw_data and isinstance(input, bytes):`
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,278 @@ async def unlock(
...
+            encoded_data = event_data
+        else:
+            if event_data is not None:
```
berndverst, line 1280: Why not combine the if else block like you did in start_workflow?

```python
encoded_data = json.dumps(input).encode("utf-8") if input is not None else bytes([])
```

berndverst: Add the try catch as above, but change the error message to `raise_workflow_event` and `event_data`

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1095,278 @@ def unlock(
...
+                headers=call.initial_metadata())
+        except RpcError as err:
+            raise DaprInternalError(err.details())
+
+    def raise_workflow_event(
+            self,
+            instance_id: str,
+            workflow_component: str,
+            event_name: str,
+            event_data: Union[Any, bytes, None] = None) -> DaprResponse:
```
berndverst:
```suggestion
            event_data: Optional[Union[Any, bytes]] = None) -> DaprResponse:
```

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1095,278 @@ def unlock(
...
+        # Warnings and input validation
+        warn('The Workflow API is an Alpha version and is subject to change.',
+             UserWarning, stacklevel=2)
+        validateNotBlankString(instance_id=instance_id,
+                               workflow_component=workflow_component,
+                               event_name=event_name)
+        if isinstance(input, bytes):
+            encoded_data = event_data
+        else:
+            if event_data is not None:
```
berndverst, line 1265: combine the following lines into the single line statement

Add try / except block as before

berndverst (changes requested):

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,301 @@ async def unlock(
...
+                headers=call.initial_metadata())
+        except grpc.aio.AioRpcError as err:
+            raise DaprInternalError(err.details())
+
+    async def raise_workflow_event(
+            self,
+            instance_id: str,
+            workflow_component: str,
+            event_name: str,
+            event_data: Union[Any, bytes, None] = None,
```
berndverst, line 1253:
```suggestion
            event_data: Optional[Union[Any, bytes]] = None,
```

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1095,302 @@ def unlock(
         return UnlockResponse(status=UnlockResponseStatus(response.status),
                               headers=call.initial_metadata())
 
+    def start_workflow(
+            self,
+            workflow_component: str,
+            workflow_name: str,
+            input: Union[Any, bytes, None] = None,
```
berndverst, line 1102:
```suggestion
            input: Optional[Union[Any, bytes]] = None,
```

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1095,302 @@ def unlock(
...
+                headers=call.initial_metadata())
+        except RpcError as err:
+            raise DaprInternalError(err.details())
+
+    def raise_workflow_event(
+            self,
+            instance_id: str,
+            workflow_component: str,
+            event_name: str,
+            event_data: Union[Any, bytes, None] = None,
```
berndverst, line 1237:
```suggestion
            event_data: Optional[Union[Any, bytes]] = None,
```

dapr/clients/grpc/client.py
berndverst: `encoded_data = json.dumps(input).encode("utf-8") if input is not None else bytes([])`

dapr/clients/grpc/client.py
```diff
@@ -1088,6 +1095,302 @@ def unlock(
...
+            encoded_data = event_data
+        else:
+            if event_data is not None:
+                try:
+                    if event_data is not None:
+                        encoded_data = json.dumps(event_data).encode("utf-8")
+                    else:
+                        event_data = bytes([])
+                except TypeError:
+                    raise DaprInternalError("Json Data is not serializable")
```
berndverst, line 1272:
```suggestion
                    raise DaprInternalError("raise_workflow_event: event_data must be JSON serializable")
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,301 @@ async def unlock(
...
+        if isinstance(input, bytes) and send_raw_bytes:
+            encoded_data = input
+        else:
+            try:
+                if input is not None:
+                    encoded_data = json.dumps(input).encode("utf-8")
+                else:
+                    encoded_data = bytes([])
+            except TypeError:
+                raise DaprInternalError("Json Data is not serializable")
```
berndverst, line 1157:
```suggestion
                raise DaprInternalError("start_workflow: input must be JSON serializable")
```

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,301 @@ async def unlock(
...
+            except ValueError as e:
+                raise DaprInternalError(f"Json serialization error: {e}")
```
berndverst, line 1159:
```suggestion
                raise DaprInternalError(f"start_workflow: JSON serialization error: {e}")
```

dapr/aio/clients/grpc/client.py
RyanLettieri: The line was too long for the linter so I split it into multiple

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,298 @@ async def unlock(
...
+            instance_id: Optional[str] = None,
+            workflow_options: Optional[Dict[str, str]] = dict(),
+            send_raw_bytes: bool = False) -> StartWorkflowResponse:
+        """Starts a workflow.
+
+            Args:
+                workflow_component (str): the name of the workflow component
+                                    that will run the workflow. e.g. `dapr`.
+                workflow_name (str): the name of the workflow that will be executed.
+                input (Union[Any, bytes, None]): the input that the workflow will receive.
```
cgillum, line 1127: Per our discussion, let's make the documentation more explicit
```suggestion
                input (Union[Any, bytes, None]): the input that the workflow will receive.
                                                 The input value will be serialized to JSON
                                                 by default. Use the send_raw_bytes param
                                                 to send unencoded binary input.
```
Please update the docs both for sync and async variants, as well as for the raise_event methods.

cgillum (approved):
LGTM

dapr/aio/clients/grpc/client.py
berndverst: This still needs updating

dapr/aio/clients/grpc/client.py
```diff
@@ -1103,6 +1110,304 @@ async def unlock(
...
```
berndverst, line 1127: update

berndverst (approved):

Issue #488, opened 2022-11-14
Change `unsubscribe_configuration` method to accept List[str]


sarsharma:
## Describe the proposal

The `unsubscribe_configuration` method currently accepts a single key to unsubscribe to 
https://github.com/dapr/python-sdk/blob/3d9b155af142fd531be95ceb0e2fccd237e630c2/dapr/clients/grpc/client.py#L976-L980 

Can it be changed to accept a list of strings to unsubscribe multiple config items in a single call

Also, the `get_configuration` method accepts a string of keys instead of `List[str]`, can the input `keys` be changed to `List[str]` to avoid confusion
https://github.com/dapr/python-sdk/blob/3d9b155af142fd531be95ceb0e2fccd237e630c2/dapr/clients/grpc/client.py#L905-L909

sarsharma:
cc : @akhilac1

berndverst:
Sure, but FYI, when you unsubscribe from one key it closes the stream and unsubscribes from all anyway. That's how the Dapr runtime works. @akhilac1 @sarsharma

berndverst:
Please provide feedback as requested @akhilac1 @sarsharma.

It's not possible to unsubscribe from a single key - the stream will be closed. If anything I feel that we should not allow specifying any keys to unsubscribe from and just call it `unsubscribe_configuration(self, store_name)`

berndverst:
Get configuration is fixed in #523 #491

sarsharma:
> Please provide feedback as requested @akhilac1 @sarsharma.
> 
> It's not possible to unsubscribe from a single key - the stream will be closed. If anything I feel that we should not allow specifying any keys to unsubscribe from and just call it `unsubscribe_configuration(self, store_name)`

Makes sense! We could change the `unsubscribe_configuration()` method to be similar to Java, C#, Go SDKs which allow unsubscribe using the subscription id instead of key items.

shivamkm07:
@berndverst can you please re-open this issue.

PR #558, opened 2023-04-20
subscribe method updated

shivamkm07:
# Description

The PR updates the subscribe method so that it takes a handler as an argument which will be executed whenever there is an update notification. Also the subscribe call returns ID in it's response as specified in sdk-spec(https://github.com/dapr/sig-sdk-spec/blob/main/spec/alpha/configuration.md). It updates the unsubscribe method to unsubscribe using ID, instead of a key. 

## Issue reference

Please reference the issue this PR will close: #488 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -963,7 +963,7 @@ class DaprGrpcClient:
         Args:
             store_name (str): the state store name to get from
             keys (str array): the keys of the key-value pairs to be gotten
-            handler(func (key, ConfigurationResponse)): the handler to be called when the config changes
+            handler(func (key, ConfigurationResponse)): the callback function to be called
             config_metadata (Dict[str, str], optional): Dapr metadata for configuration
 
         Returns:
--- a/dapr/clients/grpc/client.py
+++ b/dapr/clients/grpc/client.py
@@ -946,8 +946,8 @@ class DaprGrpcClient:
             self,
             store_name: str,
             keys: List[str],
-            config_metadata: Optional[Dict[str, str]] = dict(),
-            handler: Callable[[Text, ConfigurationResponse], None] = None) -> Text:
+            handler: Callable[[Text, ConfigurationResponse], None],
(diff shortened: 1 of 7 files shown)


berndverst (changes requested):
There are remaining linter issues, and the example E2E test hasn't been updated for the changed output. The `Readme.md` in the example folder acts as integration test - you can see the expected output lines there which must be modified.

Please make sure to run `tox -e type` `tox -e flake8` `tox -e py311` (or 39 or whatever - depends on the the Python version you are using) to help you catch type, linter and unit test problems.

You can run the examples integration test via `tox -e examples` , to just run the configuration API example, navigate to the examples folder and then run `./validate.sh configuration` (this assumes you have mechanical-markdown installed - via pip3 install mechanical-markdown)

shivamkm07:
> There are remaining linter issues, and the example E2E test hasn't been updated for the changed output. The `Readme.md` in the example folder acts as integration test - you can see the expected output lines there which must be modified.
> 
> Please make sure to run `tox -e type` `tox -e flake8` `tox -e py311` (or 39 or whatever - depends on the the Python version you are using) to help you catch type, linter and unit test problems.
> 
> You can run the examples integration test via `tox -e examples` , to just run the configuration API example, navigate to the examples folder and then run `./validate.sh configuration` (this assumes you have mechanical-markdown installed - via pip3 install mechanical-markdown)

Resolved

dapr/clients/grpc/client.py
```diff
@@ -942,11 +942,12 @@ def get_configuration(
             items=response.items,
             headers=call.initial_metadata())
 
-    async def subscribe_configuration(
```
berndverst, line 945: This was async before for a long time (way before I introduced an async client this week) - but watching a configuration synchronously doesn't make a ton of sense.

Making this synchronous but also handling the handler pattern are breaking changes - however that should be ok given the component wasn't stable.

berndverst (approved):

PR #565, opened 2023-05-18
Fix: do not expect QUERY_API in Redis state store capabilities

ItalyPaleAle:
The component doesn't report it supports Query API unless it's connected to a server with RedisJSON, which is not the default behavior


artursouza (approved):

ItalyPaleAle:
Note that tests in this PR will fail until dapr/dapr#6342 is merged. WHich needs this PR to see tests pass, in a chicken-and-egg problem :)

berndverst (approved):

Issue #569, opened 2023-05-25
WorkflowRuntime not initializing with correct host/port


DeepanshuA:
## Expected Behavior

When creating a Workflow App, user should be able to select a host/port as per their wish.
WorkflowRuntime should have an option to be initialized with correct host/port, just like DaprWorkflowClient.

## Actual Behavior

WorkflowRuntime expects 4001, which is a default port in underlying durabletask-python library.

## Steps to Reproduce the Problem

- Use demo_workflow example.
- Start dapr with some other port like 50005 and provide same port in the App.
- App will start complaining with sidecar not available on 4001.

## Release Note

RELEASE NOTE:

berndverst:
Not going to add this one to the milestone (and hence to release notes) because it fixes something else which is unreleased.

PR #570, opened 2023-05-25
Add when_any and fix grpc worker initialization

DeepanshuA:
# Description

This PR adds when_any/when_all support and also provides option to initialize WorkflowRuntime/TaskHubGrpcWorker with correct host/port.

## Issue reference

Please reference the issue this PR will close: #569 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/examples/demo_workflow/app.py
+++ b/examples/demo_workflow/app.py
@@ -33,11 +33,7 @@ def hello_act(ctx: WorkflowActivityContext, input):
 
 def main():
     host = settings.DAPR_RUNTIME_HOST
-    if host is None:
-        host = "localhost"
     port = settings.DAPR_GRPC_PORT
-    if port is None:
-        port = "4001"
     workflowRuntime = WorkflowRuntime(host, port)
     workflowRuntime.register_workflow(hello_world_wf)
     workflowRuntime.register_activity(hello_act)
--- a/examples/demo_workflow/app.py
+++ b/examples/demo_workflow/app.py
@@ -32,17 +32,16 @@ def hello_act(ctx: WorkflowActivityContext, input):
     print(f'New counter value is: {counter}!', flush=True)
 
 def main():
-    workflowRuntime = WorkflowRuntime()
-    workflowRuntime.register_workflow(hello_world_wf)
-    workflowRuntime.register_activity(hello_act)
-    workflowRuntime.start()
-
     host = settings.DAPR_RUNTIME_HOST
     if host is None:
         host = "localhost"
     port = settings.DAPR_GRPC_PORT
     if port is None:
         port = "4001"
+    workflowRuntime = WorkflowRuntime(host, port)
+    workflowRuntime.register_workflow(hello_world_wf)
+    workflowRuntime.register_activity(hello_act)
+    workflowRuntime.start()
 
     workflow_client = DaprWorkflowClient(host, port)
     print("==========Start Counter Increase as per Input:==========")
--- a/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
+++ b/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
@@ -15,13 +15,13 @@ limitations under the License.
 
 from __future__ import annotations
 from datetime import datetime
-from typing import Any, TypeVar, Union
-
-from dapr.ext.workflow.util import getAddress
+from typing import Any, Optional, TypeVar
 
 from durabletask import client
+
 from dapr.ext.workflow.workflow_state import WorkflowState
 from dapr.ext.workflow.workflow_context import Workflow
+from dapr.ext.workflow.util import getAddress
 
 T = TypeVar('T')
 TInput = TypeVar('TInput')
@@ -37,15 +37,15 @@ class DaprWorkflowClient:
        This client is intended to be used by workflow application, not by general purpose
        application.
     """
-    def __init__(self, host: Union[str, None] = None, port: Union[str, None] = None):
+    def __init__(self, host: Optional[str] = None, port: Optional[str] = None):
         address = getAddress(host, port)
         self.__obj = client.TaskHubGrpcClient(host_address=address)
 
     def schedule_new_workflow(self,
                               workflow: Workflow, *,
-                              input: Union[TInput, None] = None,
-                              instance_id: Union[str, None] = None,
-                              start_at: Union[datetime, None] = None) -> str:
+                              input: Optional[TInput] = None,
+                              instance_id: Optional[str] = None,
+                              start_at: Optional[datetime] = None) -> str:
         """Schedules a new workflow instance for execution.
 
         Args:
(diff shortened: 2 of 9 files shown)


examples/demo_workflow/app.py
```diff
@@ -32,17 +32,16 @@ def hello_act(ctx: WorkflowActivityContext, input):
...
-    workflowRuntime.register_workflow(hello_world_wf)
-    workflowRuntime.register_activity(hello_act)
-    workflowRuntime.start()
-
     host = settings.DAPR_RUNTIME_HOST
     if host is None:
         host = "localhost"
     port = settings.DAPR_GRPC_PORT
     if port is None:
         port = "4001"
```
cgillum, line 40: I think it's a little strange to set these default values because `settings.DAPR_RUNTIME_HOST` and `settings.DAPR_GRPC_PORT` will never be `None`, right?

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
```diff
@@ -67,3 +68,78 @@ def wait_for_external_event(self, name: str) -> task.Task:
...
+
+    def get_tasks(self) -> List[task.Task]:
+        return self._tasks
+
+    @abstractmethod
+    def on_child_completed(self, task: task.Task[T]):
+        pass
+
+
+class WhenAllTask(CompositeTask[List[T]]):
```
cgillum, line 95: This works, but I wonder if we're creating more work than necessary here? Could we just have `when_all` and `when_any` internally call the Durable Task versions of these same functions and return the resulting tasks directly to the user? I'm not sure if there's a lot of value in wrapping the Task return types from the durable SDK like we do for other APIs.

Alternatively, would it make sense to just have this `WhenAllTask` derive directly from the Durable Task `WhenAllTask` without reimplementing its logic?

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -28,8 +29,9 @@ class WorkflowRuntime:
     """WorkflowRuntime is the entry point for registering workflows and activities.
     """
 
-    def __init__(self):
-        self.__worker = worker.TaskHubGrpcWorker()
+    def __init__(self, host: Union[str, None] = None, port: Union[str, None] = None):
```
cgillum, line 32: nit: should we use `Option` instead of `Union` here?
```suggestion
    def __init__(self, host: Option[str] = None, port: Option[str] = None):
```

examples/demo_workflow/app.py
berndverst: Correct they will never be none. They load the environment variable or otherwise fall back to the default values.

Do not set those here please!

examples/demo_workflow/app.py
```diff
@@ -32,17 +32,16 @@ def hello_act(ctx: WorkflowActivityContext, input):
...
-    workflowRuntime = WorkflowRuntime()
-    workflowRuntime.register_workflow(hello_world_wf)
-    workflowRuntime.register_activity(hello_act)
-    workflowRuntime.start()
-
     host = settings.DAPR_RUNTIME_HOST
     if host is None:
         host = "localhost"
     port = settings.DAPR_GRPC_PORT
     if port is None:
```
berndverst, line 39: Remove None check as it does not apply.

berndverst (commented):
I really do not like `when_all` - this is not a pattern used elsewhere in Python and we should not introduce this C#-ism.

We should use naming and conventions set forth by `asyncio`. Otherwise please point me at a *popular* Python native workflow library and we can consider using that naming convention.

Why don't we use `asyncio` here in general?

Please take a look at `asyncio.gather()` which returns a future that can be awaited which only returns when all tasks and/or coroutines provided to gather complete.

As far as your `when_any` is concerned, this is done in `asyncio` using `asyncio.wait(iterable_of_tasks_or_awaitables, return_when=FIRST_COMPLETED)`

And for additional inspiration you can look at the Python 3.11 only TaskGroup concept -- https://docs.python.org/3/library/asyncio-task.html#id5

I feel like we are reinventing the wheel here... @DeepanshuA @cgillum

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
berndverst: `Option` is not a thing anyway, it's called `Optional` and you must import `from typing import Optional`

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
```diff
@@ -16,7 +16,8 @@
 from __future__ import annotations
 from datetime import datetime
 from typing import Any, TypeVar, Union
-from dapr.conf import settings
+
+from dapr.ext.workflow.util import getAddress
```
berndverst, line 20: please organize these imports

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
berndverst: Yes please use Optional instead of Union[something, None]

ext/dapr-ext-workflow/dapr/ext/workflow/util.py
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from typing import Union
+from dapr.conf import settings
+
+
+def getAddress(host: Union[str, None] = None, port: Union[str, None] = None) -> str:
```
berndverst, line 20: Use `Optional` instead of Union when creating a union of something and `None`

DeepanshuA: Done - used `None` instead of `Union of sth and None`

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
DeepanshuA: Done - please check now - followed:
```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.

You should put a blank line between each group of imports.
```

ref: https://peps.python.org/pep-0008/

examples/demo_workflow/app.py
DeepanshuA: Done

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
DeepanshuA: I have simply wrapped `when_all` and `when_any` from DTF-python.
I haven't wrapped return types - that doesn't sound useful.

berndverst (approved):

PR #579, opened 2023-07-03
Fixed: do not expect content-length header

ItalyPaleAle:
The HTTP service in multiple places expected a Content-Length header or it would refuse to parse the body.

This breaks streaming (see [test failures](https://github.com/dapr/dapr/actions/runs/5446419269/jobs/9907126825)), since in case the data is streamed it's not possible to know the content's byte length in advance. The Content-Length header is not required per the HTTP specs, so we should not enforce its presence.

This fixes the issue by not relying on the Content-Length header's presence.


ItalyPaleAle:
It looks like the "fix" I added broke something else (?). And I'm not good enough with Python to understand what's broken :(

PR #580, opened 2023-07-05
[docs] Reformat index page

hhunter-ms:
# Description

Attempt to make the SDK index pages consistent. Reformatted Python index page to be consistent with .NET index page.

## Issue reference

PR will close: dapr/docs#3600

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


msfussell (commented):
Can you add Workflow onto this page, with a link to the workflow example? Same as on the .NET page?

msfussell (dismissed):
LGTM

daprdocs/content/en/python-sdk-docs/_index.md
```diff
@@ -9,46 +9,109 @@ no_list: true
...
+| [Bindings tutorial](https://github.com/dapr/quickstarts/tree/master/tutorials/bindings) | See how Dapr Python SDK works alongside other Dapr SDKs to enable bindings. |
+| [Distributed Calculator tutorial](https://github.com/dapr/quickstarts/tree/master/tutorials/distributed-calculator/python) | Use the Dapr Python SDK to handle method invocation and state persistent capabilities. |
+| [Hello World tutorial](https://github.com/dapr/quickstarts/tree/master/tutorials/hello-world) | Learn how to get Dapr up and running locally on your machine with the Python SDK. |
+| [Hello Kubernetes tutorial](https://github.com/dapr/quickstarts/tree/master/tutorials/hello-kubernetes) | Get up and running with the Dapr Python SDK in a Kubernetes cluster. |
+| [Observability tutorial](https://github.com/dapr/quickstarts/tree/master/tutorials/observability) | Explore Dapr's metric collection, tracing, logging and health check capabilities using the Python SDK. |
+| [Pub/sub tutorial](https://github.com/dapr/quickstarts/tree/master/tutorials/pub-sub) | See how Dapr Python SDK works alongside other Dapr SDKs to enable pub/sub applications. |
+
+
+## Available packages
```
berndverst, line 67: This is incorrect.

In Python the Actor SDK is part of the main SDK, it's not a separate SDK. Please update these cards to match the information you removed above.

We have the main SDK (which includes Actors - no need to call that out specifically), we have the following Extension SDKs: Workflows (yes this is an extension SDK - the package is named `dapr-ext-workflow`), Flask, FastAPI and gRPC. Those latter 3 are primarily for handling PubSub subscriptions, Input Bindings etc in your app more easily.

berndverst (changes requested):

daprdocs/content/en/python-sdk-docs/_index.md
hhunter-ms: @berndverst working on this - another Q: should the `python-workflow.md` doc be moved in the TOC to sit under "Extensions"?

berndverst: Technically yes - but that's only because today we have the main SDK and extensions for receiving requests from the sidecar (which arguably the Workflow authoring SDK does too). We can probably leave it in place where it is to make it more easily discoverable however.

By the way, can you update the docs to say that the Workflow SDK is Alpha? Once we version it as 1.0 we will consider it stable (and hence not make potential breaking changes)

daprdocs/content/en/python-sdk-docs/_index.md
```diff
@@ -9,46 +9,142 @@ no_list: true
...
 {{% /codetab %}}
 
 {{< /tabs >}}
 
+
+## Available packages
+
+Once you've installed the main Python SDK, you can:
+- Import additional packages as part of the main SDK package
+- Install extensions on top of the main Python SDK
```
berndverst, line 49: Non-blocking: Technically when you install the extensions it will install the main SDK anyway because that's a dependency of those extensions.

daprdocs/content/en/python-sdk-docs/python-client.md
```diff
@@ -8,16 +8,18 @@ description: How to get up and running with the Dapr Python SDK
...
+{{% /alert %}}
+
+## Prerequisites
+
+[Install the Dapr Python module]({{< ref "python-sdk-docs/_index.md" >}}) before getting started.
 
 ## Import the client package
 
-The dapr package contains the `DaprClient` which will be used to create and use a client.
+The Dapr package contains the `DaprClient`, which is used to create and use a client.
```
berndverst, line 22: Maybe we should use lowercase dapr to match the import. By using the code formatting we can indicate this is intentional.

```suggestion
The `dapr` package contains the `DaprClient`, which is used to create and use a client.
```

berndverst (commented):
Looks good - super minor feedback. There is a paragraph I would remove under the Main SDK section.

berndverst (approved):

PR #607, opened 2023-09-29
Fixes workflow example

ItalyPaleAle:
Also reverts #606

The fix for now is to add a "sleep(2)" to await for the WF engine to be ready. It's perhaps the best we can do for now, until we add support for healthchecks before starting the WF client.


berndverst (approved):
Super hacky, and not a good solution - but at least it's only 2 seconds.

Still, the real problem lies in runtime IMO

Independently the SDK should check sidecar health in the future.

Issue #615, opened 2023-10-11
call_child_workflow breaks due to orchestrator name being lost


bkc:
Using dapr-ext-workflow version 0.2.0

## Expected Behavior

This code should execute the designated workflow:

```python
raw_download_result = yield ctx.call_child_workflow(
            workflow=files_com_wf.files_com_download_file_wf,
            input=path,
            instance_id=None,
 )
```

## Actual Behavior

An exception is raised

```python
== APP ==   File "/workspaces/wms-integration/wms_integration/dapr/workflows/combined_order_flow.py", line 28, in dl_and_import_order_file_wf
== APP ==     raw_download_result = yield ctx.call_child_workflow(
== APP ==                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
== APP == durabletask.task.TaskFailedError: Sub-orchestration task #1 failed: A 'wf' orchestrator was not registered.
== APP == 2023-10-11 20:37:41.667 durabletask-worker INFO: b42571ae3e5e49f7a66b1445832a5345: Orchestration completed with status: FAILED
```

## Steps to Reproduce the Problem

Try to call a child workflow whose function name is not `wf`

## Release Note

RELEASE NOTE: **FIX** Bug in runtime.

PR #616, opened 2023-10-11
change call_child_workflow to pass correct wf name to durabletask worker

bkc:
# Description

durabletask.worker uses `orchestrator.__name__` [here](https://github.com/microsoft/durabletask-python/blob/c9990973a7cc18b2db1219d129961cede9eb3948/durabletask/worker.py#L387)  

but when dapr_workflow_context wraps the real workflow function in `wf`, the correct orchestrator name is lost.

This bug results in the following exception:

```python
durabletask.task.TaskFailedError: Sub-orchestration task #1 failed: A 'wf' orchestrator was not registered.
```

when called as

```python
raw_download_result = yield ctx.call_child_workflow(
            workflow=files_com_wf.files_com_download_file_wf,
            input=path,
            instance_id=None,
        )
```

This proposed fix is not pretty, but it works.

## Issue reference

Please reference the issue this PR will close: #615

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

There aren't any existing tests or even any examples that use call_child_workflow, I'm sorry I do not have time to write tests or documentation at this time.

### DIFF
--- a/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
+++ b/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
@@ -62,7 +62,7 @@ class DaprWorkflowContext(WorkflowContext):
             daprWfContext = DaprWorkflowContext(ctx)
             return workflow(daprWfContext, inp)
         # copy workflow name so durabletask.worker can find the orchestrator in its registry
-        wf.__name__ = workflow.__name__ 
+        wf.__name__ = workflow.__name__
         return self.__obj.call_sub_orchestrator(wf, input=input, instance_id=instance_id)
 
     def wait_for_external_event(self, name: str) -> task.Task:
--- a/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
+++ b/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
@@ -61,7 +61,8 @@ class DaprWorkflowContext(WorkflowContext):
         def wf(ctx: task.OrchestrationContext, inp: TInput):
             daprWfContext = DaprWorkflowContext(ctx)
             return workflow(daprWfContext, inp)
-        wf.__name__ = workflow.__name__ # copy workflow name so durabletask.worker can find the orchestrator in its registry
+        # copy workflow name so durabletask.worker can find the orchestrator in its registry
+        wf.__name__ = workflow.__name__ 
         return self.__obj.call_sub_orchestrator(wf, input=input, instance_id=instance_id)
 
     def wait_for_external_event(self, name: str) -> task.Task:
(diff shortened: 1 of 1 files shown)


ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
```diff
@@ -61,6 +61,7 @@ def call_child_workflow(self, workflow: Workflow, *,
         def wf(ctx: task.OrchestrationContext, inp: TInput):
             daprWfContext = DaprWorkflowContext(ctx)
             return workflow(daprWfContext, inp)
+        wf.__name__ = workflow.__name__ # copy workflow name so durabletask.worker can find the orchestrator in its registry
         return self.__obj.call_sub_orchestrator(wf, input=input, instance_id=instance_id)
```
cgillum, line 65: While hacky, this feels like the right fix given the current limitations of the durabletask-python library. Ideally we add a parameter to `call_sub_orchestrator` that allows overriding the name so that callers don't need to be aware of the internal implementation details.

@DeepanshuA thoughts?

cgillum (approved):

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
berndverst: I'll merge this and you can consider a better solution in a future release.

berndverst (dismissed):

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
berndverst: Comment line is too long. Please see Python style guide ;)

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
```diff
@@ -61,6 +61,7 @@ def call_child_workflow(self, workflow: Workflow, *,
         def wf(ctx: task.OrchestrationContext, inp: TInput):
             daprWfContext = DaprWorkflowContext(ctx)
             return workflow(daprWfContext, inp)
+        wf.__name__ = workflow.__name__ # copy workflow name so durabletask.worker can find the orchestrator in its registry
```
berndverst, line 64:
```suggestion
        # copy workflow name so durabletask.worker can find the orchestrator in its registry
        wf.__name__ = workflow.__name__ 
```

berndverst (approved):

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
```diff
@@ -61,6 +61,8 @@ def call_child_workflow(self, workflow: Workflow, *,
         def wf(ctx: task.OrchestrationContext, inp: TInput):
             daprWfContext = DaprWorkflowContext(ctx)
             return workflow(daprWfContext, inp)
+        # copy workflow name so durabletask.worker can find the orchestrator in its registry
+        wf.__name__ = workflow.__name__
```
berndverst, line 65:
```suggestion
        wf.__name__ = workflow.__name__
```

berndverst (approved):

PR #618, opened 2023-10-16
Updated endpoint parsing

elena-kolevska:
# Description
This PR updates the python-sdk so it satisfies the requirements for gRPC endpoint parsing added in https://github.com/dapr/proposals/pull/40.

It also adds support for all URIs that the [gRPC name resolution](https://github.com/grpc/grpc/blob/master/doc/naming.md) document defines, including dns links with authority, unix paths with absolute and relative sockets, unix-abstract and virtual sockets.

Backwards compatibility has been maintained for endpoints where tls was inferred based on the scheme `https`.

## Issue reference
https://github.com/dapr/python-sdk/issues/619

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/dapr/conf/helpers.py
+++ b/dapr/conf/helpers.py
@@ -1,5 +1,4 @@
 from urllib.parse import urlparse, parse_qs
-from typing import Optional
 
 
 class URIParseConfig:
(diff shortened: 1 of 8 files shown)


dapr/conf/helpers.py
```diff
@@ -1,56 +1,131 @@
...
+        if self.hostname.count(":") == 7:
+            # IPv6 address
+            hostname = f"[{hostname}]"
+        return hostname
+
+    def get_endpoint(self) -> str:
+        scheme = self.get_scheme()
+        port = "" if len(self.get_port()) == 0 else f":{self.port}"
+
+        if scheme == "unix":
```
artursouza, line 60: Is there a warning if the user uses the deprecated "https" or "http" schemas?

elena-kolevska: Thanks. I added it today.

tests/conf/helpers_test.py
```diff
@@ -1,144 +1,157 @@
...
+            {"url": "inv-scheme:myhost:5000", "error": True},
+        ]
 
         for testcase in testcases:
-            o = parse_endpoint(testcase["endpoint"])
-
-            self.assertEqual(testcase["scheme"], o[0])
-            self.assertEqual(testcase["host"], o[1])
-            self.assertEqual(testcase["port"], o[2])
+            # if testcase["error"]:
```
berndverst, line 140: Can you remove the commented out lines please if these are not necessary?

berndverst (commented):
Please create a follow-up PR to remove these orphaned comments.

berndverst (approved):

elena-kolevska:
Thanks @berndverst. Good catch, I was using those prints to generate the matrix in [this proposal](https://github.com/dapr/proposals/pull/41).
Fixed in https://github.com/dapr/python-sdk/pull/633

dapr/conf/helpers.py
```diff
@@ -1,56 +1,183 @@
...
+from warnings import warn
+from urllib.parse import urlparse, parse_qs, ParseResult
+
+
+class URIParseConfig:
+    DEFAULT_SCHEME = "dns"
+    DEFAULT_HOSTNAME = "localhost"
+    DEFAULT_PORT = 443
+    DEFAULT_AUTHORITY = ""
+    ACCEPTED_SCHEMES = ["dns", "unix", "unix-abstract", "vsock", "http", "https", "grpc", "grpcs"]
```
lrascao, line 10: cc @elena-kolevska 
something's off here,  both `grpc` and `grpcs` are not present in the [naming resolution doc](https://github.com/grpc/grpc/blob/master/doc/naming.md) and are also missing in the go-sdk: https://github.com/dapr/go-sdk/blob/main/client/internal/parse.go#L160

elena-kolevska: Fixed in this PR: https://github.com/dapr/python-sdk/pull/700

PR #636, opened 2023-11-15
Workflow: Add timedelta to create_timer

### LINKED ISSUE #625: [Workflow] DaprWorkflowContext create_timer method should accept a timedelta
## Describe the proposal

In `DaprWorkflowContext` the `create_timer` method is defined as follows:

```python
def create_timer(self, fire_at: datetime) -> task.Task:
```

However, it would be more convenient to also support `timedelta` as a parameter type. Note that the undelrying Durable Task SDK already supports this:

```python
def create_timer(self, fire_at: Union[datetime, timedelta]) -> Task:
```

The workaround is that developers must do the following:

```python
ctx.create_timer(ctx.current_utc_datetime + SOME_TIME_DELTA)
```
This workaround is problematic because it can tempt users to violate code constraint rules if they accidentally use `datetime.now()`, which is non-deterministic, instead of `ctx.current_utc_datetime`.

Ideally, users could simply write the following:

```python
ctx.create_timer(SOME_TIME_DELTA)
```

berndverst:
# Description

Fixes #625

### DIFF
--- a/examples/workflow/monitor.py
+++ b/examples/workflow/monitor.py
@@ -37,7 +37,7 @@ def status_monitor_workflow(ctx: wf.DaprWorkflowContext, job: JobStatus):
             ctx.call_activity(send_alert, input=f"Job '{job.job_id}' is unhealthy!")
         next_sleep_interval = 5  # check more frequently when unhealthy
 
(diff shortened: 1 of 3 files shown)


cgillum (approved):
:shipit:

PR #641, opened 2023-11-23
add an outbound health check

### LINKED ISSUE #611: Check outbound healthz endpoint on sidecar wait
## Feature request

Similar to dotnet sdk check the outbound healthz endpoint for checking if sidecar is ready. 
https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Client/DaprClientGrpc.cs#L1790

Even if the port is open, Dapr might not be ready yet. 

## Current behavior
Python only waits to check if the port is available and socket connection can be created. 
https://github.com/dapr/python-sdk/blob/master/dapr/clients/grpc/client.py#L1431

RELEASE NOTE:

mukundansundar:
# Description

closes #611 

## Issue reference

Please reference the issue this PR will close: #611 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/dapr/clients/__init__.py
+++ b/dapr/clients/__init__.py
@@ -19,6 +19,7 @@ from warnings import warn
 from dapr.clients.base import DaprActorClientBase
 from dapr.clients.exceptions import DaprInternalError, ERROR_CODE_UNKNOWN
 from dapr.clients.grpc.client import DaprGrpcClient, MetadataTuple, InvokeMethodResponse
+from dapr.clients.http.helpers import DaprHealthClient
 from dapr.clients.http.dapr_actor_http_client import DaprActorHttpClient
 from dapr.clients.http.dapr_invocation_http_client import DaprInvocationHttpClient
 from dapr.conf import settings
@@ -72,6 +73,7 @@ class DaprClient(DaprGrpcClient):
         """
         super().__init__(address, interceptors, max_grpc_message_length)
         self.invocation_client = None
+        self.helath_client = DaprHealthClient(timeout=http_timeout_seconds)
 
         invocation_protocol = settings.DAPR_API_METHOD_INVOCATION_PROTOCOL.upper()
 
@@ -173,3 +175,27 @@ class DaprClient(DaprGrpcClient):
         else:
             raise NotImplementedError(
                 'Please use `dapr.aio.clients.DaprClient` for async invocation')
+
+    def wait(self, timeout_s: float):
+        """Wait for the client to become ready. If the client is already ready, this
+        method returns immediately.
+
+        Args:
+            timeout_s (float): The maximum time to wait in seconds.
+
+        Throws:
+            DaprInternalError: if the timeout expires.
+        """
+        self.helath_client.wait(int(timeout_s))
+
+    async def wait_async(self, timeout_s: float):
+        """Wait for the client to become ready. If the client is already ready, this
+        method returns immediately.
+
+        Args:
+            timeout_s (float): The maximum time to wait in seconds.
+
+        Throws:
+            DaprInternalError: if the timeout expires.
+        """
+        await self.helath_client.wait_async(int(timeout_s))
--- a/dapr/clients/http/helpers.py
+++ b/dapr/clients/http/helpers.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+
+"""
+Copyright 2023 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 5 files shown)


dapr/clients/__init__.py
```diff
@@ -173,3 +175,27 @@ async def invoke_method_async(
         else:
             raise NotImplementedError(
                 'Please use `dapr.aio.clients.DaprClient` for async invocation')
+
+    def wait(self, timeout_s: float):
```
mukundansundar, line 179: This is float because DaprGrpcClient already has a method wait as float, I can remove that method if that would be better....

dapr/clients/__init__.py
```diff
@@ -173,3 +175,27 @@ async def invoke_method_async(
...
+
+        Args:
+            timeout_s (float): The maximum time to wait in seconds.
+
+        Throws:
+            DaprInternalError: if the timeout expires.
+        """
+        self.helath_client.wait(int(timeout_s))
+
+    async def wait_async(self, timeout_s: float):
```
mukundansundar, line 191: This is float because DaprGrpcClient already has a method wait as float, I can remove that method if that would be better....

dapr/clients/http/helpers.py
```diff
@@ -0,0 +1,81 @@
...
+
+from dapr.clients.http.client import DaprHttpClient, USER_AGENT_HEADER, DAPR_USER_AGENT
+from dapr.serializers import DefaultJSONSerializer
+
+
+class DaprHealthClient:
+    """Dapr Health Client"""
+
+    def __init__(self, timeout: Optional[int] = 60):
+        self._client = DaprHttpClient(DefaultJSONSerializer(), timeout, None, None)
```
mukundansundar, line 29: is there a need to pass in `address` from DaprClient initialization?
Both http and grpc might not have the same endpoint address.

mukundansundar (commented):
I wanted to split out health check apis from the service invocation api, so that even when service invocation protocol is chosen as grpc, http is still used as the health check protocol.

dapr/clients/http/helpers.py
berndverst: That's unrelated to this PR. We can open an issue for discussion in the future. I don't think we need to tackle that until someone has a concrete need for it. The HTTP Client is only used by Actors and by Service Invocation, though using the SDK for service invocation is no longer the recommended way. Perhaps you can take a look how the DotNet SDK handles this and compare.

berndverst (changes requested):
Instead of creating another instance of DaprHTTPClient and also the Serializer - why not a simple static utility method that uses `urllib` ? I would much prefer that.

That method could then be imported within the wait methods and used there.

Allocating an entire Dapr HTTP Client with Serializer just feels inefficient from a memory perspective.

berndverst:
Your change broke the configuration API example by the way

```
		== APP == Traceback (most recent call last):
		== APP ==   File "configuration.py", line 47, in <module>
		== APP ==     asyncio.run(executeConfiguration())
		== APP ==   File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/asyncio/runners.py", line 44, in run
		== APP ==     return loop.run_until_complete(main)
		== APP ==   File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
		== APP ==     return future.result()
		== APP ==   File "configuration.py", line 25, in executeConfiguration
		== APP ==     d.wait(20)
		== APP ==   File "/home/runner/work/python-sdk/python-sdk/dapr/clients/__init__.py", line 189, in wait
		== APP ==     self.helath_client.wait(int(timeout_s))
		== APP ==   File "/home/runner/work/python-sdk/python-sdk/dapr/clients/http/helpers.py", line 81, in wait
		== APP ==     loop.run_until_complete(awaitable)
		== APP ==   File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/asyncio/base_events.py", line 592, in run_until_complete
		== APP ==     self._check_running()
		== APP ==   File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/asyncio/base_events.py", line 552, in _check_running
		== APP ==     raise RuntimeError('This event loop is already running')
		== APP == RuntimeError: This event loop is already running
		== APP == sys:1: RuntimeWarning: coroutine 'DaprHealthClient.wait_async' was never awaited
```

dapr/clients/__init__.py
```diff
@@ -72,6 +73,7 @@ def __init__(
         """
         super().__init__(address, interceptors, max_grpc_message_length)
         self.invocation_client = None
+        self.helath_client = DaprHealthClient(timeout=http_timeout_seconds)
```
berndverst, line 76:
```suggestion
        self.health_client = DaprHealthClient(timeout=http_timeout_seconds)
```

berndverst:
Update on this?

Issue #626, opened 2023-10-27
workflow Logs enhancements - Structured logging and right log levels


rabollin:
Improve the debuggability of Dapr Workflows by (a) identifying gaps in logging and improving them, and (b) making the error reporting consistent across.

**Explore structured logging** like `TIME LEVEL WORKFLOW_ID SOURCE MESSAGE` where SOURCE can be Dapr Client or DTF (can have subcategories which we MAY consider including worker/client/task - mostly not required from Python SDK POV). Also requires setting up logger formatter in DTF Clients for each SDK. 

 **Logs grouping** for Starting the DTF worker, successful connection to gRPC server etc.. Currently it shows just this one statement.  

 _== APP - WorkflowConsoleApp == INFO: Durable Task worker is connecting to sidecar at 127.0.0.1:61659._ 

 **Workflow Execution:**  The logs are capturing only very superficial as Workflow started, waiting for activity as shown here (logs from Python SDK). Capture Workflow Name, Workflow ID too in the messages below. 

== APP - order-processor == 2023-10-20 18:24:48.276 durabletask-worker INFO: Starting gRPC worker that connects to 127.0.0.1:61723 

== APP - order-processor == 2023-10-20 18:24:48.285 durabletask-worker INFO: Successfully connected to 127.0.0.1:61723. Waiting for work items... 

== APP - order-processor ==   start_resp = daprClient.start_workflow(workflow_component=workflow_component, 

== APP - order-processor == Starting order workflow, purchasing 10 of cars 

**Workflow Activity level logging**: The logs below show as waiting for tasks and events, but nothing beyond that. Should capture Parent ID (Workflow ID), Task ID, Task name, Task status, event ID, event status as relevant in each log message.  

== APP - order-processor == 2023-10-20 18:24:52.332 durabletask-worker INFO: 57044fb1-5f7b-4912-ad6e-f87595cb6ec2: Waiting for 1 task(s) and 0 event(s). 

18:24:52.402 durabletask-worker INFO: 57044fb1-5f7b-4912-ad6e-f87595cb6ec2: Waiting for 1 task(s) and 1 event(s). 

**Categorization of log levels** for Workflow logs for ease – Warning, Info & Debug 

Debug logs to contain more finer details about activities:   Make sure that SDKs support disabling Debug logs to not bombard app logs with too many logs

shivamkm07:
Following are some of the logging enhancements which should improve Workflow debuggability, w.r.t python-sdk 

[...]

PR #645, opened 2023-12-05
Add logs to Dapr Workflows

shivamkm07:
# Description

This PR does the following:
1. Adds the support for adding logger in Dapr
2. Adds logs in dapr-ext-workflow
3. Passes the logger options to DTF-python to enable logging

## Issue reference

Please reference the issue this PR will close: #626 
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


berndverst (changes requested):
Unless you are ready to refactor all other logging code for the rest of the Python SDKs please move the logger into a sub package of `dapr-ext-workflow` for now. (`ext/dapr-ext-workflow` sub folder)

berndverst:
My strong preference is for self contained PRs. So if you'd like to introduce a logger to be used for all Python SDKs, I would like to see that done in a single PR. That's why I am asking for the logger code to be moved to the workflow extension directory for now. As long as the code is self contained there I will approve :)

In the future we could still migrate that to be at the top level of the repo, but at that point I'd like the same PR to contain the switch to that logger throughout all SDKs.

shivamkm07:
> My strong preference is for self contained PRs. So if you'd like to introduce a logger to be used for all Python SDKs, I would like to see that done in a single PR. That's why I am asking for the logger code to be moved to the workflow extension directory for now. As long as the code is self contained there I will approve :)
> 
> In the future we could still migrate that to be at the top level of the repo, but at that point I'd like the same PR to contain the switch to that logger throughout all SDKs.

Done

dapr/logger/options.py
```diff
@@ -0,0 +1,50 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from typing import Union
+import logging
+
+
+class LoggerOptions:
```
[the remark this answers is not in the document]
shivamkm07: @mukundansundar Added logger class. Please review

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
```diff
@@ -49,26 +55,33 @@ def is_replaying(self) -> bool:
...
         def wf(ctx: task.OrchestrationContext, inp: TInput):
-            daprWfContext = DaprWorkflowContext(ctx)
+            daprWfContext = DaprWorkflowContext(ctx, self._logger.get_options())
             return workflow(daprWfContext, inp)
         # copy workflow name so durabletask.worker can find the orchestrator in its registry
         wf.__name__ = workflow.__name__
         return self.__obj.call_sub_orchestrator(wf, input=input, instance_id=instance_id)
 
     def wait_for_external_event(self, name: str) -> task.Task:
+        self._logger.debug(f'instance_id {self.instance_id} wait_for_external_event: {name}')
```
mukundansundar, line 80: if we are going to follow a k:v approach, I would prefer to use a utility method with a particular format. The other option is we have full sentences in logs similar to the other logs in python-sdk.

ext/dapr-ext-workflow/dapr/ext/workflow/logger/options.py
```diff
@@ -0,0 +1,40 @@
...
+        # Set default log level to INFO if none is provided
+        if log_level is None:
+            log_level = logging.INFO
+        # Add a default log handler if none is provided
+        if log_handler is None:
+            log_handler = logging.StreamHandler()
+        # Set a default log formatter if none is provided
+        if log_formatter is None:
+            log_formatter = logging.Formatter(
+                fmt="%(asctime)s.%(msecs)03d %(name)s %(levelname)s: %(message)s",
```
mukundansundar, line 36: is this format similar to the default format for python logs? 

If not, is there a possibility that when using workflow SDK, there will be two different format of logs from the SDK?

mukundansundar (commented):
overall lgtm

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
shivamkm07: The formatter is added as part of `LoggerOptions`. It's set to `TIME SCOPE LOG_LEVEL: LOG_MESSAGE` by default.

ext/dapr-ext-workflow/dapr/ext/workflow/logger/options.py
shivamkm07: For now, there is no logging capability in python-sdk yet

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
mukundansundar: but then you are formatting the message as `k:v` pairs ... if so would suggest to add a simple utility method for that

mukundansundar (approved):

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
shivamkm07: Moved to sentences

mukundansundar (approved):

berndverst (approved):

Issue #623, opened 2023-10-26
allow optional explicit naming of workflow and activity names


bkcsfi:
## Describe the proposal

It would be helpful to allow users to explicitly declare the name of an activity or workflow when they are registered, and the name of a called activity, as an alternative to using the activity or workflow function `__name__`.

Consider an application that can download and process files from various sources, A, B and C. In the future, additional sources will be added.

Here's one way to lay out the code.

```
/src
  /interface
     /A
        /activities.py
     /B
        /activities.py
     /C
        /activities.py
     /D
        /activities.py
```

Suppose each source has the same nominal activity `get_input_file`

I think it would be cleaner to use these activities this way:

```python
from interface import A, B, C, D
...
input_module = A if input_type == 'A' else B
file_result = yield ctx.call_activity(input_module.get_input_file)
...
```

Currently the activity name will be taken from the function name, there's no way to have the same activity in different modules.

Therefore I have to write code like this instead:

```python
from interface import A, B, C, D
...
if input_type == 'A':
    file_result = yield ctx.call_activity(A.get_input_file_A)
else:
    file_result = yield ctx.call_activity(B.get_input_file_B)
...
```

If an optional name can be explicitly passed to call_activity, register_activity and register_workflow, then users could  manage  complicated workflows in a tidier way.

for example, I would write a decorator to assign an `__registered_name` attribute to my activity functions.

Then I would use a wrapper function to call activities, something like:

```python

def call_activity(ctx, activity, *arg):
   yield ctx.call_activity(activity.__registered_name, *arg)

...

from interface import A, B, C, D
...
input_module = A if input_type == 'A' else B
file_result = yield call_activity(ctx,input_module.get_input_file)
...
```

mukundansundar:
PR #651 should address this issue.

Issue #635, opened 2023-11-07
Python decorators for Workflow


mukundansundar:
## Describe the proposal
The current Python SDK implementation of Workflow is structured and provides a set of functions to operate the workflow of your code/ long running scenarios.  
e.g.: `schedule_new_workflow()`, `get_workflow_state()`, `wait_for_workflow_start()`

The proposal is to leverage the strengths of Python programming model, to simplify the developers’ efforts further, by introducing decorators pattern to add additional features/ functionality to existing functions without modifying the original structure of the code.  

## Proposed Solution

The Workflow lifecycle is involved with multiple steps as shown below and we have multiple functions for each of these steps. 
<img width="201" alt="image" src="https://github.com/dapr/python-sdk/assets/65565396/474c0ee5-d3b6-47e2-bde7-95d6d6f2dd54">

Currently in Python workflow SDK, we have workflow function definition, activity functions definition followed by registering of the workflow and activity function against the `WorkflowRuntime` instance. 
For now, there is only one `WorkflowRuntime` instance created which interacts with the Dapr sidecar is shown below.

```python
wfr = WorkflowRuntime(host, port)
```
Given that workflow instance(`wfr`), we can register multiple workflow functions against it as well as multiple activity functions against it.  

- Decorators at **Workflow** level
  - `@wfr.workflow.register(name)` – Register the workflow with DAPR side car. `name` is related to the GitHub [issue](https://github.com/dapr/python-sdk/issues/623).
  - `@wfr.worflow.serializer(serializer)/@wfr.workflow.deserializer(deserializer)` – to set serialization/ deserialization logic with workflow input/output
  - `@wfr.workflow.retry()`– Resiliency, to apply the retries on Workflow. 
    - In durabletask-python library, there are options to set RetryPolicy for an activity, this functionality is to extend it to the workflow level and be able to set a default policy for any run of a particular workflow.
- Decorators at **Activity** level
  - `@wfr.activity.register(name)` – Register an Activity with workflow runtime. `name` is related to the GitHub issue.
  - `@wfr.activity.serializer(serializer)/@wfr.activity.deserializer(deserializer)` – to set serialization/ deserialization logic with activity input/output. 
  - `@wfr.activity.retry()` - Resiliency, to apply the retries on Activity. Currently an activity can be called with a `RetryPolicy`, based on the code in `durabletask-python` library. This is to be able to set a default retry policy for the activity even if a retry policy is not specified at the call site. 

[...]

PR #651, opened 2023-12-12
Initial code for decorators and optional naming of workflows and activities

mukundansundar:
# Description

- When a decorator is added for an activity or workflow, it is registered against the instance through which the decorator is accessed.
- For optional naming, I followed a slightly modified suggestion made in this [issue](https://github.com/dapr/python-sdk/issues/623). I add an `_alternate_name` for each `workflow`, `activity` function/method and when a new name a given, the `alternate_name` which is the user given name takes precedence over the `__name__` of the function/method. 
- Pending TODOs:
  - [x] Adding unit tests as needed for workflow/activity decorator and optional name feature. 
  - [ ] The serialize/deserializer addition will be done as a separate PR. A separate issue will be created to track it. 

Following through with the discussion in #575, I have added partial validation to the `chaining` and `fan out fan in` examples alone as the other two require user input and take a longer time to complete also. 

@berndverst  and @cgillum  For now I have left the `register_workflow` and the `register_activity` methods as is without any changes, so users can use either a decorator patter or use `register_workflow` and `register_activity` for now. 

So if we continue to support both, there has to be the case that any parameter or functionality added as part of the decorator will in essence be added as part of the `register_*` methods, that is the code flow today also. 

## Issue reference

part of #635 
closes https://github.com/dapr/python-sdk/issues/623

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


ext/dapr-ext-workflow/tests/test_workflow_runtime.py
```diff
@@ -34,20 +34,56 @@ def add_named_activity(self, name: str, fn):
...
+        assert listOrchestrators == wanted_orchestrator
+        assert client_wf._registered_name == self.mock_client_wf.__name__
+
+        client_activity = (self.runtime_options.activity())(self.mock_client_activity)
+        wanted_activity = [self.mock_client_activity.__name__]
+        assert listActivities == wanted_activity
+        assert client_activity._registered_name == self.mock_client_activity.__name__
+
+    def test_both_decorator_and_register(self):
+        client_wf = (self.runtime_options.workflow(name="test_wf"))(self.mock_client_wf)
```
mukundansundar, line 70: When called with the `@` operator, the function object is replaced in place with the wrapped function only.

bkcsfi:
Hi,

I like where this is going but I have some suggestions that would make it easier for my use case.

I think it would be helpful to offer a decorator function that sets the function `_registered_name` attribute independent of having to instantiate a WorkflowRuntime. 

All of our code explicitly passes an already instantiated  WorkflowRuntime instance from the top-most level to registration functions in lower-level modules. This way we only instantiate a WorkflowRuntime once, at the top level, and not on every module import. Being able to set `_registered_name` separately from the registration process would match our current architecture.


Then  `register_workflow()` and `register_activity()` could be adjusted to detect when the passed function already has a `_registered_name` attribute and use that if the `name` parameter had not been set. If neither `name` is passed nor `_registered_name` is found, then default to function `__name__`

mukundansundar:
> Hi,
> 
> I like where this is going but I have some suggestions that would make it easier for my use case.
> 
> I think it would be helpful to offer a decorator function that sets the function `_registered_name` attribute independent of having to instantiate a WorkflowRuntime.
> 
> All of our code explicitly passes an already instantiated WorkflowRuntime instance from the top-most level to registration functions in lower-level modules. This way we only instantiate a WorkflowRuntime once, at the top level, and not on every module import. Being able to set `_registered_name` separately from the registration process would match our current architecture.
> 
> Then `register_workflow()` and `register_activity()` could be adjusted to detect when the passed function already has a `_registered_name` attribute and use that if the `name` parameter had not been set. If neither `name` is passed nor `_registered_name` is found, then default to function `__name__`

I believe that the name a function registers with for a workflow should be provided only during the register call or be the name of the function itself. I do not think we should add a separate decorator for adding a `_registrered_name` attribute.

Instead is there any specific scenario where the optional name cannot be set with the `register_workflow()` or `register_activity()` method itself?

mukundansundar:
@bkcsfi I have modified to add an `alternate_name` decorator which adds an `_alternate_name` attribute to the function which can be used instead of `__name__` ... Please let us know if this works for your use case.

examples/workflow/fan_out_fan_in.py
```diff
@@ -59,4 +57,4 @@ def process_results(ctx, final_result: int):
     print(f'Workflow started. Instance ID: {instance_id}')
     state = wf_client.wait_for_workflow_completion(instance_id)
 
-    workflowRuntime.shutdown()
+    wfr.shutdown()
```
cgillum, line 60: This can probably be done in another PR if you'd prefer, but one of the improvements we discussed was making it so that the runtime can be stopped automatically if we allow the `start()` method to be invoked using a [`with`](https://peps.python.org/pep-0343/) block.

examples/workflow/README.md
```diff
@@ -22,10 +22,20 @@ Each of the examples in this directory can be run directly from the command line
 ### Task Chaining
 
 This example demonstrates how to chain "activity" tasks together in a workflow. You can run this sample using the following command:
+<!--STEP
```
cgillum, line 25: I appreciate you adding the mechanical markdown to this example!

examples/workflow/fan_out_fan_in.py
```diff
@@ -14,7 +14,9 @@
 from typing import List
 import dapr.ext.workflow as wf
 
+wfr = wf.WorkflowRuntime("localhost", "50001")
```
cgillum, line 17: Should we instead rely on the default values for this? Otherwise there could be problems if someone runs dapr and sets a non-standard gRPC port.

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -80,3 +107,141 @@ def start(self):
...
+        This example shows how to register a workflow function with a name:
+
+                from dapr.ext.workflow import WorkflowRuntime
+                wfr = WorkflowRuntime()
+
+                @wfr.workflow(name="add")
+                def add(ctx, x: int, y: int) -> int:
+                    return x + y
+
+        This example shows how to register a workflow function without a name:
```
cgillum, line 123: It's not that it registers it without a name, but that it uses the function name as the workflow name.

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -80,3 +107,141 @@ def start(self):
     def shutdown(self):
         """Stops the listening for work items on a background thread."""
         self.__worker.stop()
+
+    def workflow(self, __fn: Workflow = None, *, name: Optional[str] = None):
+        """Decorator to register a workflow function.
+
+        This example shows how to register a workflow function with a name:
```
cgillum, line 114:
```suggestion
        This example shows how to register a workflow function with an explicit name:
```

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -80,3 +107,141 @@ def start(self):
...
+                    from dapr.ext.workflow import WorkflowRuntime
+                    wfr = WorkflowRuntime()
+
+                    @wfr.workflow
+                    def add(ctx, x: int, y: int) -> int:
+                        return x + y
+
+        Args:
+            name (Optional[str], optional): Name to identify the workflow function as in
+            the workflow runtime. Defaults to None.
```
cgillum, line 134:
```suggestion
            the workflow runtime. Defaults to the name of the function.
```

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
```diff
@@ -53,6 +53,10 @@ def create_timer(self, fire_at: Union[datetime, timedelta]) -> task.Task:
 
     def call_activity(self, activity: Callable[[WorkflowActivityContext, TInput], TOutput], *,
                       input: TInput = None) -> task.Task[TOutput]:
+        if hasattr(activity, '_alternate_name'):
+            return self.__obj.call_activity(activity=activity.__dict__['_alternate_name'],
+                                            input=input)
+        # this return should ideally never execute
```
cgillum, line 59: Why is this "ideal"? Isn't this the case where the user chose not to use the decorator, which we plan to allow?

cgillum (commented):
A few small comments.

examples/workflow/fan_out_fan_in.py
mukundansundar: Will leave it for another PR.

examples/workflow/fan_out_fan_in.py
mukundansundar: Will change it.

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
mukundansundar: changed it to `alternate` name

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
mukundansundar: In this case, it does default to `None`, but is then overridden by the function name.

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_context.py
mukundansundar: No in both scenarios, the attribute `_alternate_name` will be set for the function being called. So ideally the if clause will always execute.

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
cgillum: I guess it depends on whether this description is intended to describe the function signature of the what the user can expect - i.e., the semantic behavior.

cgillum (approved):
I'm good with these changes. 👍🏽

bkcsfi:
@mukundansundar Hi thank you for revising the decorator and register functions. The changes meet all my needs.

Do you have an existing issue or discussion for `The serialize/deserializer addition will be done as a separate PR. A separate issue will be created to track it.` ?   

I am using some ugly approaches to try to keep activity and workflow  'input deseralization' and 'output serialization' tied to activity and workflow function declarations, rather than forcing the caller to do that work. 

I would appreciate seeing what you're planning there, as well as maybe getting durable task to `send()` deserialized activity results to the workflow..

Thanks,
-Brad

mukundansundar:
> @mukundansundar Hi thank you for revising the decorator and register functions. The changes meet all my needs.
> 
> Do you have an existing issue or discussion for `The serialize/deserializer addition will be done as a separate PR. A separate issue will be created to track it.` ?
> 
> I am using some ugly approaches to try to keep activity and workflow 'input deseralization' and 'output serialization' tied to activity and workflow function declarations, rather than forcing the caller to do that work.
> 
> I would appreciate seeing what you're planning there, as well as maybe getting durable task to `send()` deserialized activity results to the workflow..
> 
> Thanks, -Brad

I will be creating an issue following a comment/discussion from you in discord.

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -50,17 +52,30 @@ def __init__(self, host: Optional[str] = None, port: Optional[str] = None):
...
             daprWfContext = DaprWorkflowContext(ctx)
             if inp is None:
                 return fn(daprWfContext)
             return fn(daprWfContext, inp)
 
-        self.__worker._registry.add_named_orchestrator(fn.__name__, orchestrationWrapper)
-
-    def register_activity(self, fn: Activity):
+        if hasattr(fn, '_workflow_registered'):
+            # whenever a workflow is registered, it also has _alternate_name attribute
```
DeepanshuA, line 64:
```suggestion
            # whenever a workflow is already registered, it also has an existing _alternate_name attribute
```

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -71,7 +86,19 @@ def activityWrapper(ctx: task.ActivityContext, inp: Optional[TInput] = None):
                 return fn(wfActivityContext)
             return fn(wfActivityContext, inp)
 
-        self.__worker._registry.add_named_activity(fn.__name__, activityWrapper)
+        if hasattr(fn, '_activity_registered'):
+            # whenever an activity is registered, it also has _alternate_name attribute
```
DeepanshuA, line 90:
```suggestion
            # whenever an activity is already registered, it also has an existing _alternate_name attribute
```

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -50,17 +52,30 @@ def __init__(self, host: Optional[str] = None, port: Optional[str] = None):
         self.__worker = worker.TaskHubGrpcWorker(host_address=uri.endpoint, metadata=metadata,
                                                  secure_channel=uri.tls)
 
-    def register_workflow(self, fn: Workflow):
+    def register_workflow(self, fn: Workflow, *, name: Optional[str] = None):
```
DeepanshuA, line 55: Is it possible to add a small doc section somewhere in the python-sdk repo itself, where-in we briefly describe that re-registering (or mixing the registration methods to register multiple times) will not succeed?

bkcsfi: I had previously suggested that registering a workflow/activity with an explicit name argument that already has an assigned _alternate_name should raise an exception.

But now I wonder if there is a legitimate use-case for overriding a previously set `_alternate name`  during the registration step.  

Is it likely that activities and workflows might be packaged into a "python library", and the consumer of that library might want to override the alternate name during the registration process? 

Perhaps to have precise control over versioning so that existing workflows won't break?

Should the register function log a warning instead of raising an exception,  and then override the previously assigned `_alternate_name` with the passed in `name`?

I do think the decorator should raise an exception if `_alternate_name` is already set on the function. 

Finally, should the `_alternate_name` attribute be renamed to `_dapr_alternate_name`  just to namespace it a bit more?

mukundansundar: I think making the change from `_alternate_name` to`_dapr_alternate_name` is fine, I will make that change. 
On the other parts, I do not think for this PR we need to allow multiple registrations of a workflow with different names. Let's open a separate feature request and discuss the pros, cons of that feature.

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
mukundansundar: changed.

ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
mukundansundar: changed

berndverst (approved):

PR #657, opened 2024-01-10
Add ruff auto fix

litan1106:
# Description

- added a standardized way to lint and auto-format code. (like the golang fmt auto formatted)

# How to use it?

- execute `tox -e ruff` to auto-format code

## Issue reference

- we discussed it in discord python-sdk channel
- 
## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


berndverst (changes requested):
This PR breaks the type check.

Locally, please run `tox -e type`, `tox -e flake8`, and `tox -e py311`

litan1106:
> This PR breaks the type check.
> 
> Locally, please run `tox -e type`, `tox -e flake8`, and `tox -e py311`

All passed now. I just turned off the ruff auto format on these two lines.
image

elena-kolevska:
Thanks for the PR @litan1106. I had been thinking about adding a auto fixer for a while too.
One question - what's the reason for having both flake and ruff?

elena-kolevska:
I understand. Do they have different behaviour though? If we can configure ruff to use the same lint rules as flake8, keeping it simple with a single tool that does both things would be nice. One problem would be if we plan on using some of the flake plugins in the future... What do you think @berndverst?

berndverst:
@elena-kolevska flake8 is a much much more common linter for Python. That's why I don't want to switch to ruff for linting :)

berndverst:
@litan1106 I force pushed to your PR - I split everything up into two commits.
1: Adding ruff configurations
2: Autoformatting everything using `tox -e ruff`
3: Adds a ruff check to CI

berndverst:
We can consider adding ruff as the linter in the future by the way. For now let us keep flake8 :)

berndverst (approved):

PR #670, opened 2024-02-07
Health check

elena-kolevska:
# Description

This PR adds a health check against Dapr's health endpoint instead of checking if the socket is open.

The check is added on all the clients' init methods (grpc, http, async grpc, actor and invocation). 

The timeout can be configured through the `DAPR_HEALTH_TIMEOUT` environment variable.

The default value for `DAPR_HEALTH_TIMEOUT` is 60 seconds.

## Issue reference

https://github.com/dapr/python-sdk/issues/611

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

## Release notes
When connecting to a remote sidecar, `DAPR_HTTP_ENDPOINT` needs to be set, even when only gRPC is used, because the health endpoint is http-only.


dapr/clients/health.py
```diff
@@ -0,0 +1,82 @@
...
+import aiohttp
+
+from dapr.clients.http.conf import DAPR_API_TOKEN_HEADER, USER_AGENT_HEADER, DAPR_USER_AGENT
+from dapr.clients.http.helpers import get_api_url
+from dapr.conf import settings
+
+HEALTHY = False
+
+
+def healthcheck(timeout_s: int = 5):
```
berndverst, line 31: Small suggestion - make the attribute simply `seconds` instead of `timeout_s`

dapr/clients/http/client.py
```diff
@@ -17,23 +17,26 @@
...
-CONTENT_TYPE_HEADER = 'content-type'
-DAPR_API_TOKEN_HEADER = 'dapr-api-token'
-USER_AGENT_HEADER = 'User-Agent'
-DAPR_USER_AGENT = f'dapr-sdk-python/{__version__}'
 
 
 class DaprHttpClient:
     """A Dapr Http API client"""
 
+    @healthcheck(5)
```
berndverst, line 39: And based on my suggestion, use this like `@healthchecks(seconds=5)` - makes it more readable :)

dapr/actor/client/proxy.py
```diff
@@ -98,7 +98,7 @@ class ActorProxy:
     communication.
     """
 
-    _default_proxy_factory = ActorProxyFactory()
```
elena-kolevska, line 101: I moved this to the instance creation because it was getting invoked on import(dependency load), which was triggering the healthcheck before the fake test servers would start.

tests/clients/test_dapr_grpc_client.py
```diff
@@ -42,19 +42,24 @@
...
+    grpc_port = 50001
+    http_port = 3500
     scheme = ''
     error = None
 
-    def setUp(self):
-        self._fake_dapr_server = FakeDaprSidecar()
-        self._fake_dapr_server.start(self.server_port)
+    @classmethod
+    def setUpClass(cls):
```
elena-kolevska, line 51: I moved the server creation in all tests to the `setUpClass` method. With this, we got a speed improvement of around 300% (from ~6m to ~1.45m).

berndverst (approved):

Issue #676, opened 2024-02-26
[FEATURE REQUEST] Implement retry and timeout policy


elena-kolevska:
## Describe the feature

Based on https://github.com/dapr/dapr/issues/6609 and https://github.com/dapr/proposals/blob/main/0010-S-sdk-resiliency.md

## Release Note
RELEASE NOTE: **ADD** Implement retry and timeout policy for both grpc and http based clients

elena-kolevska:
Completed in https://github.com/dapr/python-sdk/pull/679

PR #679, opened 2024-02-27
Retry and Timeout policies for grpc and http

elena-kolevska:
## Description

This PR adds the possibility for users to configure Retry and Timeout policies for both grpc and http based clients.
The retry is configurable through the `DAPR_API_MAX_RETRIES` environment variable and is picked up automatically for both the grpc and http client, not requiring any code changes.
I left the possibility for a more fine-grained control over the retry mechanism (the client can receive a RetryPolicy object), but I'm not sure if that's something people would need. Let's discuss it.

## Issue reference

https://github.com/dapr/python-sdk/issues/676

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


elena-kolevska:
The lint tests are currently failing because of a change introduced in Ruff 0.3 which is incompatible with our flake8 set-up. We should solve this issue in a separate PR.

berndverst:
Is this PR ready @elena-kolevska ?

elena-kolevska:
It's almost ready, I just wanted to confirm a few things with you:

- Right now the code will retry if the env variable is set, with no additional code changes. I also made it possible for users to tune other retry-related parameters (like initial_backoff, max_backoff, retryable_http_status_codes for ex.). Do you think the need ffor it justifies the added code complexity, and heavier constructors?
- For now, I only added the retry wrapper to one method, the `save_state` one. Once we confirm this is the right approach, I'll update all methods in the class.
- When I first started work on this PR I added a timeout interceptor that would be applied to all calls. Looking at it now, I think it makes more sense to be able to specify a timeout per method. Some things are expected to take longer than others, and control should be more granular.

berndverst:
> It's almost ready, I just wanted to confirm a few things with you:
> 
> * Right now the code will retry if the env variable is set, with no additional code changes. I also made it possible for users to tune other retry-related parameters (like initial_backoff, max_backoff, retryable_http_status_codes for ex.). Do you think the need ffor it justifies the added code complexity, and heavier constructors?
> * For now, I only added the retry wrapper to one method, the `save_state` one. Once we confirm this is the right approach, I'll update all methods in the class.

I think that's probably fine, but why wouldn't we instead add this as a decorator on the save_state method instead? Would that be possible? Right now you are not actually using a decorator but instead a method that takes another method as the argument (which is very similar to how decorators work). I just find decorators a lot cleaner. Keep in mind that you can always have the decorator intercept arbitrary **kwargs and then receive whatever properties you need to tweak retry behavior, without having to modify the method signature of the function.

> * When I first started work on this PR I added a timeout interceptor that would be applied to all calls. Looking at it now, I think it makes more sense to be able to specify a timeout per method. Some things are expected to take longer than others, and control should be more granular. What do you think?

How do you plan for users to be able to tweak the individual method timeouts? As optional arguments during each of the methods? That would work if you use decorators, because decorators can intercept arguments.

We should not add any new parameters to methods or the constructor - everything should be optional to not cause a breaking change and we should choose either sane defaults or retain the existing behavior, but then document how this can be set.

Let's also make sure using an env variable will not be the only way to control this behavior by the way.

elena-kolevska:
Yeah, decorators were my first try, and I need to go back and remember why we decided it wasn't the best option. I'll get back to you on it. 
In the meantime, on the topic of timeouts. Yeah, we would indeed have to change every method's signature, which is tedious (it would be an optional parameter, of course), but I think it's the right way to do it because it reflects how you would use it in real life. It also follows the gRPC way, where method calls receive timeout arguments (`with_call`).

elena-kolevska:
I remembered why we're not using decorators: when we retry we would have to call again the whole function that prepares the request, instead of just the call, so it's better to keep the code more performant and use a single function.

I also went ahead and updated all the calls and marked the PR as ready to review 🙏

berndverst (approved):

PR #683, opened 2024-03-06
Brings back a docs link that was removed because it wasn’t online yet

elena-kolevska:
Brings back the link we removed because it wasn’t online yet


elena-kolevska:
@berndverst this link wasn't online until today and we changed it so the examples tests wouldn't fail. 
The documents just got published, so I'm adding back the link.

berndverst:
@elena-kolevska linter ruff autoformatter is failing. Can you update?

elena-kolevska:
Since version 0.3.0 the ruff autoformatter conflicts with flake8. I think this is something we have to decide and send a  separate PR for.

berndverst:
I pinned 0.2.2 now and updated your PR

PR #700, opened 2024-04-24
Removes grpc and grpcs from the accepted schemes list

elena-kolevska:
# Description

`grpc` and `grpcs` are not valid schemes, as per the [gRPC name resolution spec](https://github.com/grpc/grpc/blob/master/doc/naming.md) and they're not included in our [test matrix](https://github.com/dapr/proposals/blob/main/20230627-S-sidecar-endpoint-tls.md#test-matrix), but they were mistakenly left over in the accepted schemes list.


marcduiker:
@holopin-bot @elena-kolevska 😁

PR #725, opened 2024-06-25
Fixes linter

elena-kolevska:
Flake8 linter passes locally, but fails in CI (with the same version). This addresses the formatting it's complaining about.


berndverst:
Fixed in another PR.

Issue #737, opened 2024-10-03
[FEATURE REQUEST] Mock actor objects for using testing


lor1113:
I would like to suggest a mock/test actor object, implemented in the Python SDK, that would allow for easy and portable unit testing of various methods, without any dependency on the Dapr sidecar being up, or other processes being active. Essentially it would be an way to create an object that has all the methods an actor has, giving you the ability to call them in tests and check the result. 

A code example of what I envisage the final result would look like would be (the specifics or names here do not particularly matter, it's just for demonstrating the idea):
```
testactor = TestActor(MyActor)
assert testactor.some_method(some_args) == what_you_expect
```

How I envisage this interacting with the various Actor features:

Actor activation / _on_activate()
This could either be automatically called on object instantiation, essentially replacing `__init__` (just as Dapr does when instantiating a new actor instance) or be left as a method for the user to call manually . I'm not really sure what method I prefer. If you keep it as a separate method every time the mock object is used one would have to manually call the function (assuming it does some important setup) which is clutter, but on the other hand, it might be useful to keep it separate.

State storage:
This would be handled entirely locally, with each mock actor object holding its own state as a dict or other dict like object. Ideally this could be passed in at object creation, and read from the object directly. A code example of what a test would look like:

```
testactor = TestActor(MyActor, initstate=some_state)
testactor.method_that_changes_state(some_args)
assert testactor.state == what_you_expect
```
the ActorStateStore object used by the mock actor would only write to this temporary in memory dict-like store, which would only last the lifetime of each object and have no persistence - obviously, if you want to test actors activating or responding to requests in an environment with a populated state store, all you need to do is pass it in or set it.

Timers and reminders:
These should not have any of their actual functionality within the mock object, since it's simply outside the scope of what the mock actors are trying to do. You would simply unit test the methods they would call with the arguments they would use (and more complex end to end testing is outside of the scope of this proposal).

There does remain the question of testing adding and removing reminders / timers. For this, I propose that the mock actor register/unregister timer/reminder methods be overrwritten to simply point to an externally accessible list/dict of `ActorReminderData` and `ActorTimerData` instances which holds all the "active" timers/reminders for each actor object, which test code can then check and query to see if timers/reminders are being added and removed correctly (They can also be removed via TTL, of course, but again I would say that testing that functionality is outside of the scope of this). An code example of a test using this would look like:

```
testactor = TestActor(MyActor)
testactor.method_that_adds_reminder(some_args)
[...]

PR #750, opened 2024-11-01
Add Actor Mocks

lor1113:
### New merge request made due to manually copying changes into a new branch being seemingly the only way to avoid weird git diff issues.

# Description

Added Mock Actor capability.

NOTE: Due to not being able to get dapr-docs running in its own dev container despite repeated attempts (I am simply not a frontend coder so i don't know how to debug all the weird errors it was spitting out), I have not been able to actually view the new documentation within a running instance of the website. I have written it nonetheless using the same templating as the others, as well as basic markdown syntax, but it would be appreciated if someone who has a develop version of the docs website working on their machine could just give it a quick look to make sure it's not somehow broken.

## Issue reference

Please reference the issue this PR will close: #737

## Checklist

* [x] Code compiles correctly
* [X] Created/updated tests


elena-kolevska (changes requested):
I'm sorry I haven't gotten to this PR yet, we're all really busy with the 1.15 release. 
I noticed you renamed one of the docs files. I think we should revert that change, because this .md is shown directly in the dapr docs: https://docs.dapr.io/developing-applications/sdks/python/python-actor/.
Also, please move the new docs you added in the existing page.

lor1113:
@elena-kolevska 

The file was renamed such that it would be able to have sub-files, like how the main "Extensions" page under the Python SDK documentation is called "index.md" (https://docs.dapr.io/developing-applications/sdks/python/python-sdk-extensions/) with the idea that the mock actor documentation could be confined to a sub-file for brevity.

Requested changes have been made regardless. It's now simply appended to the end of the current actor documentation.

dapr/actor/runtime/mock_actor.py
```diff
@@ -0,0 +1,123 @@
...
+        class SomeActorInterface(ActorInterface):
+            @actor_method(name="method")
+            async def set_state(self, data: dict) -> None:
+
+        class SomeActor(Actor, SomeActorInterface):
+            async def set_state(self, data: dict) -> None:
+                await self._state_manager.set_state('state', data)
+                await self._state_manager.save_state()
+
+        mock_actor = create_mock_actor(SomeActor)
```
elena-kolevska, line 42:
```suggestion
        mock_actor = create_mock_actor(SomeActor, "actor_1")
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
 
 ## Sample
 
-Visit [this page](https://github.com/dapr/python-sdk/tree/release-1.0/examples/demo_actor) for a runnable actor sample.
\ No newline at end of file
+Visit [this page](https://github.com/dapr/python-sdk/tree/release-1.0/examples/demo_actor) for a runnable actor sample.
```
elena-kolevska, line 59:
```suggestion
Visit [this page](https://github.com/dapr/python-sdk/tree/v1.14.0/examples/demo_actor) for a runnable actor sample.
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+
+
+## Mock Actor Testing
+
+The Dapr Python SDK provides the ability to create mock actors to unit test your actor methods and how they interact with the actor state.
```
elena-kolevska, line 64:
```suggestion
The Dapr Python SDK provides the ability to create mock actors to unit test your actor methods and see how they interact with the actor state.
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+    async def save_state(self, data) -> None:
+        await self._state_manager.set_state('state', data)
+        await self._state_manager.save_state()
+
+mock_actor = create_mock_actor(MyActor, "id")
+
+await mock_actor.save_state(5)
+assert mockactor._state_manager._mock_state == 5 #True
+```
+Mock actors work by passing your actor class as well as the actor id (str) into the function create_mock_actor, which returns an instance of the actor with a bunch of the internal actor methods overwritten, such that instead of attempting to interact with Dapr to save state, manage timers, etc it instead only uses local variables.
```
elena-kolevska, line 82:
```suggestion
Mock actors are created by passing your actor class and an actor ID (a string) to the create_mock_actor function. This function returns an instance of the actor with many internal methods overridden. Instead of interacting with Dapr for tasks like saving state or managing timers, the mock actor uses in-memory state to simulate these behaviors.
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+* **_state_manager._mock_state()**
+A [str, object] dict where all the actor state is stored. Any variable saved via _state_manager.save_state(key, value), or any other statemanager method is stored in the dict as that key, value combo. Any value loaded via try_get_state or any other statemanager method is taken from this dict.
+
+* **_state_manager._mock_timers()**
+A [str, ActorTimerData] dict which holds the active actor timers. Any actor method which would add or remove a timer adds or pops the appropriate ActorTimerData object from this dict.
+
+* **_state_manager._mock_reminders()**
+A [str, ActorReminderData] dict which holds the active actor reminders. Any actor method which would add or remove a timer adds or pops the appropriate ActorReminderData object from this dict.
+
+**Note: The timers and reminders will never actually trigger. The dictionaries exist only so methods that should add or remove timers/reminders can be tested. If you need to test the callbacks they should activate, you should call them directly with the appropriate values:**
```
elena-kolevska, line 94:
```suggestion
This state can be accessed through the following variables:
- **_state_manager._mock_state()**  
A `[str, object]` dict where all the actor state is stored. Any variable saved via `_state_manager.save_state(key, value)`, or any other statemanager method is stored in the dict as that key, value pair. Any value loaded via `try_get_state` or any other statemanager method is taken from this dict.

- **_state_manager._mock_timers()**  
A `[str, ActorTimerData]` dict which holds the active actor timers. Any actor method which would add or remove a timer adds or pops the appropriate `ActorTimerData` object from this dict.

- **_state_manager._mock_reminders()**  
A [str, ActorReminderData] dict which holds the active actor reminders. Any actor method which would add or remove a timer adds or pops the appropriate ActorReminderData object from this dict.

**Note: The timers and reminders will never actually trigger. The dictionaries exist only so methods that should add or remove timers/reminders can be tested. If you need to test the callbacks they should activate, you should call them directly with the appropriate values:**
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+```
+result = await mock_actor.recieve_reminder(name, state, due_time, period, _ttl)
+# Test the result directly or test for side effects (like changing state) by querying _state_manager._mock_state
```
elena-kolevska, line 97:
```suggestion
# Test the result directly or test for side effects (like changing state) by querying `_state_manager._mock_state`
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+```
+
+### Usage and Limitations
+
+**The \_on\_activate method will not be called automatically the way it is when Dapr initializes a new Actor instance. You should call it manually as needed as part of your tests.**
+
+The \_\_init\_\_, register_timer, unregister_timer, register_reminder, unregister_reminder methods are all overwritten by the MockActor class that gets applied as a mixin via create_mock_actor. If your actor itself overwrites these methods, those modifications will themselves be overwritten and the actor will likely not behave as you expect.
+
+*note: \_\_init\_\_ is a special case where you are expected to define it as*
```
elena-kolevska, line 106:
```suggestion
### Usage and Limitations

**The `_on_activate` method will not be called automatically the way it is when Dapr initializes a new Actor instance. You should call it manually as needed as part of your tests.**

The `__init__`, `register_timer`, `unregister_timer`, `register_reminder`, `unregister_reminder` methods are all overwritten by the MockActor class that gets applied as a mixin via `create_mock_actor`. If your actor itself overwrites these methods, those modifications will themselves be overwritten and the actor will likely not behave as you expect.

*note: `__init__` is a special case where you are expected to define it as*
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+
+### Type Hinting
+
+Because of Python's lack of a unified method for type hinting type intersections (see: [python/typing #213](https://github.com/python/typing/issues/213)), type hinting is unfortunately mostly broken with Mock Actors. The return type is type hinted as "instance of Actor subclass T" when it should really be type hinted as "instance of MockActor subclass T" or "instance of type intersection [Actor subclass T, MockActor]" (where, it is worth noting, MockActor is itself a subclass of Actor).
+
+This means that, for example, if you hover over ```mockactor._state_manager``` in a code editor, it will come up as an instance of ActorStateManager (instead of MockStateManager), and various IDE helper functions (like VSCode's ```Go to Definition```, which will bring you to the definition of ActorStateManager instead of MockStateManager) won't work properly.
+
+For now, this issue is unfixable, so it's merely something to be noted because of the confusion it might cause. If in the future it becomes possible to accurately type hint cases like this feel free to open an issue about implementing it.
```
elena-kolevska, line 123:
```suggestion
### Type Hinting

Because of Python's lack of a unified method for type hinting type intersections (see: [python/typing #213](https://github.com/python/typing/issues/213)), type hinting unfortunately doesn't work with Mock Actors. The return type is type hinted as "instance of Actor subclass T" when it should really be type hinted as "instance of MockActor subclass T" or "instance of type intersection `[Actor subclass T, MockActor]`" (where, it is worth noting, `MockActor` is itself a subclass of `Actor`).

This means that, for example, if you hover over `mockactor._state_manager` in a code editor, it will come up as an instance of ActorStateManager (instead of MockStateManager), and various IDE helper functions (like VSCode's `Go to Definition`, which will bring you to the definition of ActorStateManager instead of MockStateManager) won't work properly.

For now, this issue is unfixable, so it's merely something to be noted because of the confusion it might cause. If in the future it becomes possible to accurately type hint cases like this feel free to open an issue about implementing it.
```

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+*note: \_\_init\_\_ is a special case where you are expected to define it as*
+```
+    def __init__(self, ctx, actor_id):
+        super().__init__(ctx, actor_id)
+```
+*Mock actors work fine with this, but if you have added any extra logic into \_\_init\_\_, it will be overwritten. It is worth noting that the correct way to apply logic on initialization is via \_on\_activate (which can also be safely used with mock actors) instead of \_\_init\_\_.*
+
+The actor _runtime_ctx variable is set to None. Obviously all the normal actor methods have been overwritten such as to not call it, but if your code itself interacts directly with _runtime_ctx, it will likely break.
+
+The actor _state_manager is overwritten with an instance of MockStateManager. This has all the same methods and functionality of the base ActorStateManager, except for using the various _mock variables for storing data instead of the _runtime_ctx. If your code implements its own custom state manager it will be overwritten and your code will likely break.
```
elena-kolevska, line 115:
```suggestion
*Mock actors work fine with this, but if you have added any extra logic into `__init__`, it will be overwritten. It is worth noting that the correct way to apply logic on initialization is via `_on_activate` (which can also be safely used with mock actors) instead of `__init__`.*

The actor `_runtime_ctx` variable is set to None. All the normal actor methods have been overwritten such as to not call it, but if your code itself interacts directly with `_runtime_ctx`, it will likely break.

The actor _state_manager is overwritten with an instance of `MockStateManager`. This has all the same methods and functionality of the base `ActorStateManager`, except for using the various `_mock` variables for storing data instead of the `_runtime_ctx`. If your code implements its own custom state manager it will be overwritten and your code will likely break.
```

dapr/actor/runtime/state_manager.py
```diff
@@ -69,6 +70,9 @@ def __init__(self, actor: 'Actor'):
         self._type_name = actor.runtime_ctx.actor_type_info.type_name
 
         self._default_state_change_tracker: Dict[str, StateMetadata] = {}
+        self._mock_state: Dict[str, Any]
```
elena-kolevska, line 73: Do we need this?

dapr/actor/runtime/mock_state_manager.py
```diff
@@ -0,0 +1,235 @@
...
+        if state_name in self._default_state_change_tracker:
+            state_metadata = self._default_state_change_tracker[state_name]
+            if state_metadata.change_kind == StateChangeKind.remove:
+                self._default_state_change_tracker[state_name] = StateMetadata(
+                    value, StateChangeKind.update
+                )
+                return True
+            return False
+        existed = state_name in self._mock_state
+        if not existed:
```
elena-kolevska, line 54:
```suggestion
        if existed:
```

elena-kolevska (changes requested):
Looks great @lor1113!
I suggested some text cleanup in the docs (thank you for writing such a detailed guide).
As you had noticed before, there's a bug in the `try_add_state` method of the state manager, and I already sent a pr here: https://github.com/dapr/python-sdk/pull/756.
Feel free to update your mock to follow the same logic.

Two requests: please add a unit test for the mock state manager and an example for using a mock in the `examples` directory, that also serves as our suite of integration tests.

Again, great work!

dapr/actor/runtime/state_manager.py
lor1113: If you don't add this, then because of the type hinting problems mentioned earlier, the _mock_state, _mock_reminders and _mock_timer variables won't be correctly type hinted by the IDE, and will show up as white unknown variables.

With those type hint stubs
image

Without

image

Yes the code would still run just fine, but it's not very nice, no?

Type hinting the MockStateManager is of no use, because (as you can see in the image) it thinks the _state_manager variable it is dealing with is an ActorStateManager instead of MockStateManager.

In my view adding these stubs to ActorStateManager is harmless (at least, I cannot think of a scenario where they might cause problems or be undesirable) so I think they're worth having.

examples/demo_actor/demo_actor/test_demo_actor.py
```diff
@@ -0,0 +1,46 @@
...
+        mockactor = create_mock_actor(DemoActor, '1')
+        self.assertFalse(mockactor._state_manager._mock_state)
+        val = await mockactor.get_my_data()
+        self.assertIsNone(val)
+
+    async def test_set_data(self):
+        mockactor = create_mock_actor(DemoActor, '1')
+        self.assertFalse(mockactor._state_manager._mock_state)
+        val = await mockactor.get_my_data()
+        self.assertIsNone(val)
```
elena-kolevska, line 23: We don't need to repeat the same check as above

examples/demo_actor/README.md
```diff
@@ -183,3 +184,47 @@ expected_stdout_lines:
...
+
+<!-- STEP
+name: Actor Tests
+background: true
+sleep: 5
+expected_stdout_lines:
+  - '----------------------------------------------------------------------'
+  - 'OK'
+timeout_seconds: 60
+-->
```
elena-kolevska, line 197:
```suggestion
<!-- STEP
name: Actor Tests
background: true
expected_stdout_lines:
  - "has_value: False"
  - "set_my_data: {'state': 5}"
  - "has_value: True"
  - "clear_my_data"
  - "has_value: False"
  - "has_value: False"
  - "set reminder to True"
  - "set reminder is done"
  - "set reminder to False"
  - "set reminder is done"
  - "has_value: False"
  - "set_my_data: {'state': 5}"
  - "has_value: True"
expected_stderr_lines:
  - "----------------------------------------------------------------------"
  - "Ran 5 tests in"
  - "OK"
output_match_mode: substring
timeout_seconds: 60
-->
```

tests/actor/test_mock_state_manager.py
```diff
@@ -0,0 +1,116 @@
...
+class ActorMockActorTests(unittest.IsolatedAsyncioTestCase):
+    def test_init_state(self):
+        mock_actor = create_mock_actor(MockTestActor, 'test')
+        state_manager = mock_actor._state_manager
+        self.assertIsInstance(state_manager, MockStateManager)
+        self.assertFalse(state_manager._mock_state)
+        self.assertFalse(state_manager._mock_reminders)
+        self.assertFalse(state_manager._mock_timers)
+
+    async def test_add_state(self):
```
elena-kolevska, line 30: I think we can merge most of these tests (the add, get, set, remove, contains...) into one single test, so we don't have a lot of repetition. Just to keep the tests snappier and infra usage lower.

elena-kolevska (changes requested):
here's the fix for the examples test and a few other suggestions. Looking through the other changes now.

dapr/actor/runtime/state_manager.py
elena-kolevska: I'm leaning more towards removing the `_mock` arguments from the main code, and rely on good documentation (which @lor1113  already did) for awareness of the properties' existence. What do you think @berndverst ?

elena-kolevska (changes requested):

examples/demo_actor/demo_actor/test_demo_actor.py
lor1113: This is resolved in latest commit.

daprdocs/content/en/python-sdk-docs/python-actor.md
```diff
@@ -56,4 +56,68 @@ async def main():
...
+```
+result = await mock_actor.recieve_reminder(name, state, due_time, period, _ttl)
+# Test the result directly or test for side effects (like changing state) by querying _state_manager._mock_state
+```
+
+### Usage and Limitations
+
+**The `_on_activate` method will not be called automatically the way it is when Dapr initializes a new Actor instance. You should call it manually as needed as part of your tests.**
+
+The `__init__`, `register_timer`, `unregister_timer`, `register_reminder`, `unregister_reminder` methods are all overwritten by the MockActor class that gets applied as a mixin via `create_mock_actor`. If your actor itself overwrites these methods, those modifications will themselves be overwritten and the actor will likely not behave as you expect.
```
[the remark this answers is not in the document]
lor1113: It's a bit hard to think about how to suggest a workaround here, as by definition if someone is overriding those methods they're already doing something pretty out-of-the-box where it's hard to suggest a generic fix. But I did add in something in the latest commit, see if you approve.

elena-kolevska (approved):

marcduiker:
@holopin-bot @lor1113 Thank you! Here's a digital badge as a small token of appreciation.

Issue #739, opened 2024-10-10
[WORKFLOW SDK FEATURE REQUEST] support set custom status


famarting:
## Describe the WORKFLOW SDK feature

to bring parity with other dapr workflow sdks

## Release Note

RELEASE NOTE:

famarting:
base implemented here https://github.com/microsoft/durabletask-python/pull/31

elena-kolevska:
Resolved in https://github.com/dapr/python-sdk/pull/757

Issue #711, opened 2024-05-08
[WORKFLOW SDK FEATURE REQUEST] Purge API


cgillum:
## Describe the WORKFLOW SDK feature

The Workflow SDK's `DaprWorkflowClient` class is missing the `purge_workflow` API. This should be added for parity with other SDKs.

### Workaround

As a workaround, the `DaprGrpcClient.purge_workflow` API can be used. However, Python developers are encouraged to use `DaprWorkflowClient` instead of `DaprGrpcClient` for managing workflows.

## Release Note

RELEASE NOTE: **ADD** Purge API for Workflow SDK

elena-kolevska:
Solved in https://github.com/dapr/python-sdk/pull/757

PR #757, opened 2024-12-03
workflows: update durabletask dependency

famarting:
# Description

Update durabletask dependency to use new dapr fork

Also implements support for purge API https://github.com/dapr/python-sdk/issues/711

Implements support for reuse id policy and set custom status from https://github.com/dapr/python-sdk/issues/739

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ext/dapr-ext-workflow/tests/test_workflow_client.py
```diff
@@ -58,6 +59,9 @@ def suspend_orchestration(self, instance_id: str):
     def resume_orchestration(self, instance_id: str):
         return mock_resume_result
 
+    def purge_workflow(self, instance_id: str, recursive: bool = True):
```
elena-kolevska, line 62:
```suggestion
    def purge_orchestration(self, instance_id: str, recursive: bool = True):
```

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
```diff
@@ -15,9 +15,11 @@
 
 from __future__ import annotations
 from datetime import datetime
-from typing import Any, Optional, TypeVar
+from typing import Any, Optional, TypeVar, Union
```
elena-kolevska, line 18:
```suggestion
from typing import Any, Optional, TypeVar
```

elena-kolevska (changes requested):
Some comments. Please run `tox -e ruff` and `tox -e flake8` to fix linter errors.

We'll also need to update the grpc library to >= 1.67.0, for compatibility with durabletask.

And finally, we'll need to update the tests in `test_dapr_workflow_context.py` to cover `set-custom_status`.  Something like 
```python
class FakeOrchestrationContext:
    def __init__(self):
        self.instance_id = mock_instance_id
        self.custom_status = None. # <- adding this
    def set_custom_status(self, custom_status):
        self.custom_status = custom_status
        
...

class DaprWorkflowContextTest(unittest.TestCase):
...
def test_workflow_context_functions(self):
...
            dapr_wf_ctx.set_custom_status(mock_status)
            assert fakeContext.custom_status == mock_status            
```

setup.cfg
```diff
@@ -26,8 +26,8 @@ include_package_data = True
 zip_safe = False
 install_requires =
     protobuf >= 4.22
-    grpcio >= 1.37.0
-    grpcio-status>=1.37.0
+    grpcio >= 1.67.0
```
elena-kolevska, line 29: @berndverst any objections on bumping the grpc library version? 
This has been updated in the durable task fork and it's a dependency here.

elena-kolevska (dismissed):
LGTM, pending @berndverst 's approval on the grpcio version bump

setup.cfg
elena-kolevska: Version 1.37.0 is very old, it's from April 2021. I think we're safe to update.

berndverst: This does not mean it would use 1.37 but that 1.37 would be possible to be used. Pip by default will always install the latest version matching the constraint. But if another library a customer is using required `grpcio==1.47.0` for example it would be possible for that specific version to be used.

In Python two versions cannot coexist unlike in other languages. So it's really really important not to force the minimum version to something recent unless this is absolutely necessary.

dev-requirements.txt
```diff
@@ -15,4 +15,4 @@ Flask>=1.1
 # needed for auto fix
 ruff===0.2.2
 # needed for dapr-ext-workflow
-durabletask>=0.1.1a1
+durabletask-dapr >= 0.2.0a3
```
berndverst, line 18: FYI it is a terrible decision that this fork made `grpcio` minimum version as high as it did because it impacts interop with lots of older libraries that pinned older versions. There are plenty of Dapr customers using outdated ML libraries that do this.

elena-kolevska: Since the dapr python sdk is the only user of durabletask-dapr, it's not problematic to reevaluate this decision. @famarting , what do you say?

setup.cfg
berndverst:
```suggestion
    grpcio >= 1.42.0
```

Short of auditing the Dapr community to figure out what libraries everyone uses and what the dependencies are pinning... here is a quick thing I identified. This is evidence of at least 1.42.0 being used for something.

https://github.com/tensorflow/tensorflow/blob/8fd1cd0826ef2b12c00540f3acc5a6e3c249f5b2/ci/official/containers/linux_arm64/devel.usertools/test.requirements.txt#L2


If as a project we want to say that we do not care about ML users - or that ML users must be on the latest official `grpcio` packages (highly unlikely because ML package authors are very slow to update packages) then we could consider increasing the minimum version of `grpcio`.

berndverst (changes requested):

setup.cfg
elena-kolevska: Unfortunately, this version is imposed by the protos used in the [durable task library](https://github.com/microsoft/durabletask-python/blob/main/durabletask/internal/orchestrator_service_pb2_grpc.py#L11).

@cgillum  and @famarting, can we regenerate the protos with an older grpcio version? I suggest we pin it to a version we agree on in the requirements file. I'll check for any CVEs and share here, so we can decide which version would make sense.

berndverst: Yes let's please do that -- you should not use the latest grpcio-tools to generate the protos -- instead use an older version for wider compatibility with more versions of grpcio.

Is this only an issue in the "fork" or also the upstream version?

elena-kolevska: It's coming from the upstream, here's the commit: https://github.com/microsoft/durabletask-python/pull/31.

famarting: if we have to update the protos, I would suggest doing it here https://github.com/dapr/durabletask-protobuf

elena-kolevska: Here's the PR: https://github.com/dapr/durabletask-python/pull/6
Once that's merged I'll publish a new version and we can update this PR.

elena-kolevska (approved):

PR #766, opened 2025-01-08
Updates protos and fixes grpc-tools for protos generation

elena-kolevska:


tools/requirements.txt
```diff
@@ -1 +1 @@
-grpcio-tools>=1.57.0
+grpcio-tools==1.57.0
```
berndverst, line 1:
```suggestion
grpcio-tools==1.62.3
```

I did some research and found that `1.62.3` is the latest version before `protobuf` 1.26.X is used which has breaking changes for Python. Let's pin this to 1.62.3 to get a newer version while maintaining compatibility with more `grpcio` versions (which use older `protobuf` versions).

berndverst: Could you use that version please and regenerate the protos? :)

berndverst (approved):

Issue #745, opened 2024-10-23
[WORKFLOW SDK BUG] Get workflow state raising exception for instances that do not exist


ajstewart:
## Expected Behavior

When I run:

```python
state = wf_client.get_workflow_state(instance_id=my_id)
```

I expect to receive a `None` response when the instance does not exist, as defined by the docs:

> The current state of the workflow instance, or None if the workflow instance does not exist.

## Actual Behavior

I am getting an Exception when the instance does not exist:

```
== APP == grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
== APP == 	status = StatusCode.UNKNOWN
== APP == 	details = "error from internal actor: no such instance exists"
== APP == 	debug_error_string = "UNKNOWN:Error received from peer ipv4:127.0.0.1:55008 {grpc_message:"error from internal actor: no such instance exists", grpc_status:2, created_time:"2024-10-23T10:46:36.084357+01:00"}"
```

## Steps to Reproduce the Problem

The following small app can reproduce by calling the endpoint 

```python
import dapr.ext.workflow as wf
import uvicorn

from fastapi import FastAPI


def hello_world(ctx: wf.WorkflowActivityContext):
    """Hello world workflow."""
    return "Hello world!"


def dummy_workflow(ctx: wf.DaprWorkflowContext):
    """Dummy workflow."""
    result = yield ctx.call_activity(hello_world)
    return result


wf_runtime = wf.WorkflowRuntime()
wf_runtime.register_workflow(dummy_workflow)
wf_runtime.register_activity(hello_world)

wf_client = wf.DaprWorkflowClient()

wf_runtime.start()


app = FastAPI()


@app.get("/workflow/{instance_id}")
def get_workflow_state(instance_id: str):
    """Get the workflow state."""
    state = wf_client.get_workflow_state(instance_id=instance_id)

    if state is None:
        return {"message": "Workflow not found"}

    return {"status": state.runtime_status}


if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=5001)
```

## Release Note

RELEASE NOTE:

ajstewart:
Not stale

alicejgibbons:
@ajstewart are you running this locally? and if so, can you share the dapr logs?

ajstewart:
@alicejgibbons 

There aren't any Dapr errors really, I'm assuming the SDK doesn't handle the not found.

Here's the end of the trace that concerns the Dapr bit:

[...]

PR #771, opened 2025-01-28
Catch error in sdk when workflow instance not found

elena-kolevska:
# Description

Catches error from daprd in the sdk when workflow instance not found, so that users don't have to do try/catch in the app.
Instead, when daprd can't find the workflow instance `get_workflow_state` will return `None`, as specified in the doc comment.

Also adds a small demo workflows examples fix.

## Issue reference

https://github.com/dapr/python-sdk/issues/745

## Checklist

* [X] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
+++ b/ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
@@ -139,7 +139,7 @@ class DaprWorkflowClient:
                 self._logger.warning(f"Workflow instance not found: {instance_id}")
                 return None
             self._logger.error(f"Unhandled RPC error while fetching workflow state: {error.code()} - {error.details()}")
-            raise error
+            raise
 
     def wait_for_workflow_start(
         self, instance_id: str, *, fetch_payloads: bool = False, timeout_in_seconds: int = 60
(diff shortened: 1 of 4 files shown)


ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
```diff
@@ -130,8 +131,17 @@ def get_workflow_state(
...
         """
-        state = self.__obj.get_orchestration_state(instance_id, fetch_payloads=fetch_payloads)
-        return WorkflowState(state) if state else None
+        try:
+            state = self.__obj.get_orchestration_state(instance_id, fetch_payloads=fetch_payloads)
+            return WorkflowState(state) if state else None
+        except RpcError as error:
+            if 'no such instance exists' in error.details():
+                self._logger.warning(f'Workflow instance not found: {instance_id}')
+                return None
```
ajstewart, line 140: Does the RpcError raised by Dapr include the `NOT_FOUND` status code? Looks like you can do a `e.code()`  when looking at a real error? I'm not very familiar with `grpc` code.

I'm wondering if it's better to check by status code rather than string matching on the message?

ajstewart (commented):
On second thought I do see the sense of wrapping the not found to a `None` return and raising anything else. Seeing if it's there or not will be the most common check I imagine.

I have a question on the detection of the not found.

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
elena-kolevska: Unfortunately, this error hasn't been converted yet to the Rich Error model and it returns the default status `UNKNOWN`, so we have to check for the status message, for now.

PR #780, opened 2025-02-13
Updates workflows docs to use the new workflows client

elena-kolevska:
...and restores sleeps in test, as it was decided not to invest in making synchronous the APIs that are getting deleted next version


msfussell (commented):
Main feedback here is that I would move Workflow section to near the top, just after Service Invocation. It is buried at the bottom and it the most important API.

msfussell (commented):
Reviewing this again, I would take the lead from .NET SDK layout. https://v1-15.docs.dapr.io/developing-applications/sdks/dotnet/
Remove the workflow from the client page and I would put the code for workflow in your PR as an example on the Dapr Workflow overview page here https://v1-15.docs.dapr.io/developing-applications/sdks/python/python-sdk-extensions/python-workflow-ext/
Keep this page as a more detailed description https://v1-15.docs.dapr.io/developing-applications/sdks/python/python-sdk-extensions/python-workflow-ext/python-workflow/
That way all the workflow is in one place

elena-kolevska:
Thanks @msfussell, I updated the PR.

msfussell (approved):
LGTM

Issue #786, opened 2025-02-24
[FEATURE REQUEST] Conversation API support


elena-kolevska:
## Describe the feature
Add support for the Dapr Conversation API

## Release Note

RELEASE NOTE: **ADD** Adds support for the Dapr Conversation API

PR #787, opened 2025-02-24
Conversation API support

elena-kolevska:
# Description

Adds support for the Dapr Conversation API.

I suffixed the `converse` method name with `alpha1`. This is open for discussion. We haven't been using the suffix in the python sdk, but we do it in Go. I believe it's safer to have it suffixed, but if we're reasonably certain the API won't break we can just use `converse`.

## Issue reference
https://github.com/dapr/python-sdk/issues/786

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


dapr/aio/clients/grpc/client.py
```diff
@@ -1711,6 +1715,65 @@ async def purge_workflow(self, instance_id: str, workflow_component: str) -> Dap
...
+    ) -> ConversationResponse:
+        """Invoke an LLM using the conversation API (Alpha).
+
+        Args:
+            name: Name of the LLM component to invoke
+            inputs: List of conversation inputs
+            context_id: Optional ID for continuing an existing chat
+            parameters: Optional custom parameters for the request
+            metadata: Optional metadata for the component
+            scrub_pii: Optional flag to scrub PII from outputs
```
marcduiker, line 1738:
```suggestion
            scrub_pii: Optional flag to scrub PII from inputs and outputs
```
Scrubbing is done on both inputs and outputs:
https://github.com/dapr/dapr/blob/955436f45f783e52c9af8c2ae32f7f82a287c39c/pkg/api/universal/conversation.go#L67
https://github.com/dapr/dapr/blob/955436f45f783e52c9af8c2ae32f7f82a287c39c/pkg/api/universal/conversation.go#L116

dapr/clients/grpc/client.py
```diff
@@ -1713,6 +1717,64 @@ def purge_workflow(self, instance_id: str, workflow_component: str) -> DaprRespo
...
+    ) -> ConversationResponse:
+        """Invoke an LLM using the conversation API (Alpha).
+
+        Args:
+            name: Name of the LLM component to invoke
+            inputs: List of conversation inputs
+            context_id: Optional ID for continuing an existing chat
+            parameters: Optional custom parameters for the request
+            metadata: Optional metadata for the component
+            scrub_pii: Optional flag to scrub PII from outputs
```
marcduiker, line 1740:
```suggestion
            scrub_pii: Optional flag to scrub PII from inputs and outputs
```

marcduiker (changes requested):
Added two suggestions for the PII scrubbing

famarting (approved):

yaron2 (approved):

PR #793, opened 2025-03-14
Improve the error handling for HTTP client so consumers can trigger appropriate behavior

passuied:
# Description

Improve the error handling for HTTP client so consumers can trigger appropriate behavior

## Issue reference

Please reference the issue this PR will close: #794

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation (n/a)

### DIFF
--- a/dapr/clients/exceptions.py
+++ b/dapr/clients/exceptions.py
@@ -33,16 +33,22 @@ class DaprInternalError(Exception):
         message: Optional[str],
         error_code: Optional[str] = ERROR_CODE_UNKNOWN,
         raw_response_bytes: Optional[bytes] = None,
+        status_code: Optional[int] = None,
+        reason: Optional[str] = None,
     ):
         self._message = message
         self._error_code = error_code
         self._raw_response_bytes = raw_response_bytes
+        self._status_code = status_code
+        self._reason = reason
 
     def as_dict(self):
         return {
             'message': self._message,
             'errorCode': self._error_code,
             'raw_response_bytes': self._raw_response_bytes,
+            'status_code': self._status_code,
+            'reason': self._reason,
         }
 
     def as_json_safe_dict(self):
--- a/dapr/clients/exceptions.py
+++ b/dapr/clients/exceptions.py
@@ -133,9 +133,7 @@ class DaprHttpError(DaprInternalError):
                 error_code = error_info.get('errorCode') or ERROR_CODE_UNKNOWN
 
         super().__init__(
-            message or f'HTTP status code: {status_code}',
-            error_code,
-            raw_response_bytes
+            message or f'HTTP status code: {status_code}', error_code, raw_response_bytes
         )
 
     @property
@@ -154,7 +152,7 @@ class DaprHttpError(DaprInternalError):
 
     def __str__(self):
         if self._error_code != ERROR_CODE_UNKNOWN:
-            return f"{self._message} (Error Code: {self._error_code}, Status Code: {self._status_code})"
+            return f'{self._message} (Error Code: {self._error_code}, Status Code: {self._status_code})'
         else:
             return f'Unknown Dapr Error. HTTP status code: {self._status_code}.'
 
--- a/dapr/clients/exceptions.py
+++ b/dapr/clients/exceptions.py
@@ -58,7 +58,20 @@ class DaprInternalError(Exception):
             )
 
         return error_dict
+    @property
+    def message(self) -> Optional[str]:
+        """Get the error message"""
+        return self._message
 
+    @property
+    def error_code(self) -> Optional[str]:
+        """Get the error code"""
+        return self._error_code
+
+    @property
+    def raw_response_bytes(self) -> Optional[bytes]:
+        """Get the raw response bytes"""
+        return self._raw_response_bytes
     def __str__(self):
         if self._error_code != ERROR_CODE_UNKNOWN:
             return f"('{self._message}', '{self._error_code}')"
--- a/dapr/clients/exceptions.py
+++ b/dapr/clients/exceptions.py
@@ -33,22 +33,16 @@ class DaprInternalError(Exception):
         message: Optional[str],
         error_code: Optional[str] = ERROR_CODE_UNKNOWN,
         raw_response_bytes: Optional[bytes] = None,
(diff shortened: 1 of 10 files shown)


elena-kolevska (changes requested):
I feel like instead of modifying the generic`DaprInternalError` class with http properties that are not always relevant, we should extend it to a http specific class, something like:
``` class DaprHttpError(DaprInternalError):```

**Comment by passuied** [comment]
> I feel like instead of modifying the generic`DaprInternalError` class with http properties that are not always relevant, we should extend it to a http specific class, something like:
> ``` class DaprHttpError(DaprInternalError):```

I like it. What about including the response object in the exception then? I was hesitating about it but didn't want to couple it.

File: dapr/clients/http/client.py [hunk 1]
```diff
@@ -106,22 +106,36 @@ async def convert_to_error(self, response: aiohttp.ClientResponse) -> DaprIntern
         try:
             error_body = await response.read()
             if (error_body is None or len(error_body) == 0) and response.status == 404:
-                return DaprInternalError('Not Found', ERROR_CODE_DOES_NOT_EXIST)
+                return DaprHttpError(
+                    error_code=ERROR_CODE_DOES_NOT_EXIST,
+                    status_code=response.status,
+                    reason=response.reason,
+                )
```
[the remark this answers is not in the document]
passuied: Unfortunately, calling `response.read()` is async so I'm not sure we can use it in a constructor... copilot is recommending using `asyncio.run()` but I'm not sure that's a good idea... I'm worried about the negative performance side effects. The best I could do is create an async factory method... Would provide the desired encapsulation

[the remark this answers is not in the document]
passuied: I have refactored the code and I have come up with a good compromise (I hope). Please let me know

dapr/clients/exceptions.py
```diff
@@ -74,6 +82,53 @@ def as_dict(self):
...
+            if error_info and isinstance(error_info, dict):
+                message = error_info.get('message')
+                error_code = error_info.get('errorCode') or ERROR_CODE_UNKNOWN
+
+        super(__class__, self).__init__(
+            message or f'HTTP status code: {status_code}', error_code, raw_response_bytes
+        )
+
+    def as_dict(self):
+        error_dict = super(__class__, self).as_dict()
```
elena-kolevska, line 120:
```suggestion
        error_dict = super().as_dict()
```

dapr/clients/exceptions.py
```diff
@@ -74,6 +82,53 @@ def as_dict(self):
...
+        error_dict['status_code'] = self._status_code
+        error_dict['reason'] = self._reason
+        return error_dict
+
+    def __str__(self):
+        if self._error_code != ERROR_CODE_UNKNOWN:
+            return super(__class__, self).__str__()
```
elena-kolevska, line 127:
```suggestion
            return f"{self._message} (Error Code: {self._error_code}, Status Code: {self._status_code})"
```

dapr/clients/exceptions.py
```diff
@@ -74,6 +82,53 @@ def as_dict(self):
...
+                pass
+                # ignore any errors during deserialization
+
+            if error_info and isinstance(error_info, dict):
+                message = error_info.get('message')
+                error_code = error_info.get('errorCode') or ERROR_CODE_UNKNOWN
+
+        super(__class__, self).__init__(
+            message or f'HTTP status code: {status_code}', error_code, raw_response_bytes
+        )
```
elena-kolevska, line 117:
```suggestion
        super().__init__(message or f'HTTP status code: {status_code}', error_code,
            raw_response_bytes)
```

dapr/clients/exceptions.py
```diff
@@ -74,6 +82,53 @@ def as_dict(self):
...
+
```
elena-kolevska, line 118:
```suggestion
    @property
    def status_code(self) -> Optional[int]:
        return self._status_code

    @property
    def reason(self) -> Optional[str]:
        return self._reason
```

dapr/clients/exceptions.py
```diff
@@ -74,6 +82,53 @@ def as_dict(self):
         return {attr: getattr(self, attr) for attr in self.__dict__}
 
 
+class DaprHttpError(DaprInternalError):
+    """DaprHttpError encapsulates all Dapr HTTP exceptions"""
```
elena-kolevska, line 86:
```suggestion
    """DaprHttpError encapsulates all Dapr HTTP exceptions

    Attributes:
        _status_code: HTTP status code
        _reason: HTTP reason phrase
    """
```

elena-kolevska (changes requested):
Yes, I think that's a great approach @passuied .
I suggested a few changes; I added property decorators on both the new http class and on the existing DaprInternalError class.
This way we can access the properties from the app:
Example:

```python
        try:
            resp = d.invoke_method(
                'invoke-receiver',
                'my-method',
                data=json.dumps(req_data),
            )
        except DaprHttpError as e:
            print(e.error_code, flush=True)
            print(e.status_code, flush=True)
            print(e.message, flush=True)
            print(e.raw_response_bytes, flush=True)
            print(e.reason, flush=True)
            print(e.as_dict(), flush=True)
```

elena-kolevska (changes requested):
One last thing before we merge this. Can we pls add a failure scenario in the `examples/invoke-simple` example, that's also used as an e2e test?

You can add something lke this:
```python
        try:
            resp = d.invoke_method(
                'non-existent',
                'my-method',
                data=json.dumps(req_data),
            )
        except DaprHttpError as e:
            print(e.error_code, flush=True)
            print(e.status_code, flush=True)
            print(e.message, flush=True)
            print(e.raw_response_bytes, flush=True)
            print(e.reason, flush=True)
            print(e.as_dict(), flush=True)
```            

And then in the [README](https://github.com/dapr/python-sdk/blob/main/examples/invoke-simple/README.md?plain=1#L48) file you can add the expected output around line 48.

elena-kolevska (approved):

elena-kolevska (approved):

marcduiker:
@holopin-bot @passuied  Thank you! Here's a digital badge as a small token of appreciation.

PR #839, opened 2025-09-13
Wait for the Dapr health check asynchronously in aio/clients/grpc/subscription.py to avoid blocking, ensuring the asyncio gRPC stream can close properly.

MingsingGo:
# Switch Dapr health check from blocking call to async call to avoid blocking the event loop in async environments

# Description

This PR addresses a critical issue in the Dapr Python SDK's async pubsub subscription functionality where the gRPC stream fails to close properly during reconnection, causing the first response after reconnection to be lost. The root cause is the synchronous health check blocking the async event loop, which prevents the _StreamRequestMixin._done_writing() method from executing properly.

## Details:
1. Stream Closing Process: The close() method calls self._stream.cancel(), which triggers the _StreamRequestMixin._consume_request_iterator() to complete and call await self._done_writing().
2. Critical Issue: The _done_writing() method in _StreamRequestMixin (line 518 in grpc/aio/call.py) calls await self._cython_call.send_receive_close() to properly close the gRPC stream. However, when DaprHealth.wait_until_ready() blocks the event loop synchronously, this crucial cleanup operation cannot complete.
3. Stream State Corruption: Without proper stream closure via send_receive_close(), the gRPC stream remains in an inconsistent state. When a new stream is created after reconnection, the first response from the outgoing_request_iterator() (line 41 in subscription.py) cannot be sent because the previous stream's cleanup was incomplete.

```python
class _StreamRequestMixin(Call):
    _metadata_sent: asyncio.Event
    _done_writing_flag: bool
    _async_request_poller: Optional[asyncio.Task]
    _request_style: _APIStyle

    def _init_stream_request_mixin(
        self, request_iterator: Optional[RequestIterableType]
    ):
        self._metadata_sent = asyncio.Event()
        self._done_writing_flag = False

        # If user passes in an async iterator, create a consumer Task.
        if request_iterator is not None:
            self._async_request_poller = self._loop.create_task(
                self._consume_request_iterator(request_iterator)
            )
            self._request_style = _APIStyle.ASYNC_GENERATOR
        else:
            self._async_request_poller = None
            self._request_style = _APIStyle.READER_WRITER

    def _raise_for_different_style(self, style: _APIStyle):
        if self._request_style is not style:
            raise cygrpc.UsageError(_API_STYLE_ERROR)

    def cancel(self) -> bool:
        if super().cancel():
            if self._async_request_poller is not None:
                self._async_request_poller.cancel()
            return True
        else:
            return False

    def _metadata_sent_observer(self):
        self._metadata_sent.set()

    async def _consume_request_iterator(
        self, request_iterator: RequestIterableType
    ) -> None:
        try:
            if inspect.isasyncgen(request_iterator) or hasattr(
                request_iterator, "__aiter__"
            ):
                async for request in request_iterator:
                    try:
                        await self._write(request)
                    except AioRpcError as rpc_error:
                        _LOGGER.debug(
                            (
                                "Exception while consuming the"
                                " request_iterator: %s"
                            ),
                            rpc_error,
                        )
                        return
            else:
                for request in request_iterator:
                    try:
                        await self._write(request)
                    except AioRpcError as rpc_error:
                        _LOGGER.debug(
                            (
                                "Exception while consuming the"
                                " request_iterator: %s"
                            ),
                            rpc_error,
                        )
                        return

            await self._done_writing()
        except:  # pylint: disable=bare-except
            # Client iterators can raise exceptions, which we should handle by
            # cancelling the RPC and logging the client's error. No exceptions
            # should escape this function.
            _LOGGER.debug(
                "Client request_iterator raised exception:\n%s",
                traceback.format_exc(),
            )
            self.cancel()

    async def _write(self, request: RequestType) -> None:
        if self.done():
            raise asyncio.InvalidStateError(_RPC_ALREADY_FINISHED_DETAILS)
        if self._done_writing_flag:
            raise asyncio.InvalidStateError(_RPC_HALF_CLOSED_DETAILS)
        if not self._metadata_sent.is_set():
            await self._metadata_sent.wait()
            if self.done():
                await self._raise_for_status()

        serialized_request = _common.serialize(
            request, self._request_serializer
        )
        try:
            await self._cython_call.send_serialized_message(serialized_request)
        except cygrpc.InternalError as err:
            self._cython_call.set_internal_error(str(err))
            await self._raise_for_status()
        except asyncio.CancelledError:
            if not self.cancelled():
                self.cancel()
            raise

    async def _done_writing(self) -> None:
        if self.done():
            # If the RPC is finished, do nothing.
            return
        if not self._done_writing_flag:
            # If the done writing is not sent before, try to send it.
            self._done_writing_flag = True
            try:
                await self._cython_call.send_receive_close()
            except asyncio.CancelledError:
                if not self.cancelled():
                    self.cancel()
                raise

    async def write(self, request: RequestType) -> None:
        self._raise_for_different_style(_APIStyle.READER_WRITER)
        await self._write(request)

    async def done_writing(self) -> None:
        """Signal peer that client is done writing.

        This method is idempotent.
        """
        self._raise_for_different_style(_APIStyle.READER_WRITER)
        await self._done_writing()

    async def wait_for_connection(self) -> None:
        await self._metadata_sent.wait()
        if self.done():
            await self._raise_for_status()
```

## Solution:
The fix makes the health check asynchronous by using loop.run_in_executor(), which:
1. Prevents blocking the async event loop during health check
2. Ensures the gRPC stream is properly closed before creating a new one

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


dapr/aio/clients/grpc/subscription.py
```diff
@@ -51,7 +51,8 @@ async def outgoing_request_iterator():
 
     async def reconnect_stream(self):
         await self.close()
-        DaprHealth.wait_until_ready()
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, DaprHealth.wait_until_ready)
```
acroca: I'm wondering if this is necessary at all. Since we are using grpc shouldn't we just rely on the way grpc manages connections instead of sending a healthcheck ourselves? What would happen if we remove this call completely? @elena-kolevska you may have more context.

MingsingGo: > I'm wondering if this is necessary at all. Since we are using grpc shouldn't we just rely on the way grpc manages connections instead of sending a healthcheck ourselves? What would happen if we remove this call completely? @elena-kolevska you may have more context.

Did you mean removing the health check here? If we don't wait for Dapr to become healthy within the reconnect_stream method, the Dapr SDK will immediately attempt to restart the gRPC subscription stream. This is likely to fail, and the resulting exception would be propagated back to the point where the application calls the next_message method.

acroca: Ah, I see this `reconnect_stream` is called when we receive a `StatusCode.UNAVAILABLE` error, meaning there's something wrong with the grpc connection.

Would it make sense to have an implementation of the `DaprHealth.wait_until_ready()` async friendly?

acroca: @StarJourneyMingsing Maybe creating a `DaprHealth` in `dapr/aio/clients` with the same interface as the existing one but using asyncio? Would that work? I think it'd be cleaner, what do you think?

MingsingGo: > @StarJourneyMingsing Maybe creating a `DaprHealth` in `dapr/aio/clients` with the same interface as the existing one but using asyncio? Would that work? I think it'd be cleaner, what do you think?

It's a good idea, I think.

MingsingGo: > @StarJourneyMingsing Maybe creating a `DaprHealth` in `dapr/aio/clients` with the same interface as the existing one but using asyncio? Would that work? I think it'd be cleaner, what do you think?

I have implemented DaprHealth in dapr/aio/clients, which asynchronously waits for Dapr to become healthy using asyncio.sleep. It works well.

acroca (dismissed):

acroca (dismissed):

dapr/aio/clients/grpc/subscription.py
```diff
@@ -66,7 +66,7 @@ async def next_message(self):
                     return None
                 return SubscriptionMessage(message.event_message)
         except AioRpcError as e:
-            if e.code() == StatusCode.UNAVAILABLE:
+            if e.code() == StatusCode.UNAVAILABLE or e.code() == StatusCode.UNKNOWN:
```
elena-kolevska, line 70: `StatusCode.UNKNOWN` shouldn't be retriable, no?

tests/clients/test_dapr_grpc_client_async_secure.py
```diff
@@ -18,14 +18,16 @@
...
+from dapr.clients.health import DaprHealth as DaprHealth
 from tests.clients.certs import replacement_get_credentials_func, replacement_get_health_context
 from tests.clients.test_dapr_grpc_client_async import DaprGrpcClientAsyncTests
 from .fake_dapr_server import FakeDaprSidecar
 from dapr.conf import settings
 
 
 DaprGrpcClientAsync.get_credentials = replacement_get_credentials_func
+DaprHealthAsync.get_ssl_context = replacement_get_health_context
 DaprHealth.get_ssl_context = replacement_get_health_context
```
elena-kolevska, line 30: Why do we still need the sync version?

elena-kolevska (changes requested):

tests/clients/test_dapr_grpc_client_async_secure.py
MingsingGo: > Why do we still need the sync version?

See dapr/clients/grpc/subscription.py: the __init__ function of DaprGrpcClientAsync invokes the synchronous `wait_until_ready`. I'm not sure whether it's a good practice to implement an asynchronous factory method to create a DaprGrpcClientAsync instance, since __init__ can only be synchronous.

dapr/aio/clients/grpc/subscription.py
MingsingGo: > `StatusCode.UNKNOWN` shouldn't be retriable, no?

Actually, the exception code is not always recognized as StatusCode.UNAVAILABLE when Dapr shuts down immediately; instead, it may appear as StatusCode.UNKNOWN, as seen in the error message previously encountered: `gRPC error while reading from stream: api server closed, Status Code: StatusCode.UNKNOWN. Attempting to reconnect...`. 

The synchronous subscription implementation also retries upon encountering StatusCode.UNKNOWN, see: [https://github.com/dapr/python-sdk/commit/e7c85cea29e64581f86a7b127d22e618a972bbdb#diff-c7dbe5c0c85056e25bd5d884ad10147a7d12f4b8e9aac52513504a171404093e](https://github.com/dapr/python-sdk/commit/e7c85cea29e64581f86a7b127d22e618a972bbdb#diff-c7dbe5c0c85056e25bd5d884ad10147a7d12f4b8e9aac52513504a171404093e) (dapr/clients/grpc/subscription.py)

dapr/aio/clients/health.py
```diff
@@ -0,0 +1,55 @@
...
+        headers = {USER_AGENT_HEADER: DAPR_USER_AGENT}
+        if settings.DAPR_API_TOKEN is not None:
+            headers[DAPR_API_TOKEN_HEADER] = settings.DAPR_API_TOKEN
+        timeout = float(settings.DAPR_HEALTH_TIMEOUT)
+
+        start = time.time()
+        while True:
+            try:
+                req = urllib.request.Request(health_url, headers=headers)
+                with urllib.request.urlopen(req, context=DaprHealth.get_ssl_context()) as response:
```
acroca: Wouldn't it be better to use an async http lib like `aiohttp`?

MingsingGo: Good point! I've already switched to using aiohttp.

dapr/aio/clients/health.py
```diff
@@ -0,0 +1,68 @@
...
+
+class DaprHealth:
+    @staticmethod
+    async def wait_until_ready():
+        warn(
+            'This method is deprecated. Use DaprHealth.wait_for_sidecar instead.',
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        await DaprHealth.wait_for_sidecar()
```
acroca: Since this is new, better not ship it with a deprecated method.

dapr/aio/clients/grpc/subscription.py
```diff
@@ -52,7 +52,7 @@ async def outgoing_request_iterator():
 
     async def reconnect_stream(self):
         await self.close()
-        DaprHealth.wait_for_sidecar()
+        await DaprHealth.wait_until_ready()
```
acroca:
```suggestion
        await DaprHealth.wait_for_sidecar()
```

tests/clients/test_dapr_grpc_client_async_secure.py
acroca: Can't we use now the async `DaprHealth` in the `DaprGrpcClientAsync` init?

dapr/aio/clients/health.py
MingsingGo: removed

dapr/aio/clients/grpc/subscription.py
MingsingGo: modified

tests/clients/test_dapr_grpc_client_async_secure.py
MingsingGo: > Can't we use now the async `DaprHealth` in the `DaprGrpcClientAsync` init?

modified. use DaprHealthAsync.get_ssl_context only in tests/clients/test_dapr_grpc_client_async_secure.py

dapr/aio/clients/grpc/client.py
```diff
@@ -90,7 +90,7 @@
     UnlockResponseStatus,
 )
 from dapr.clients.grpc._state import StateItem, StateOptions
-from dapr.clients.health import DaprHealth
+from dapr.aio.clients.health import DaprHealth
```
acroca: Sorry for my previous comment, but since we're calling the health check from the `__init__`, we can't make it asynchronously because `__init__` doesn't support async. So I'm afraid we need to stick with the synchronous health check here unfortunately.
It'd be good to have some sort of `.connect` method that we can call outside of the `__init__` and would support asynchronous calls, but there's no such a thing right now, and I don't think we want to make a big change for this...
So, can you bring back the sync call please?

MingsingGo: Got it—thanks for the clarification! I've reverted to the synchronous health check.

acroca (approved):

elena-kolevska (approved):

marcduiker:
@holopin-bot @MingsingGo sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #841, opened 2025-09-17
Adds support for interceptors and concurrency_options arguments in the workflow engine

acroca:
# Description

Adds a way to configure `concurrency_options` and `interceptors` in the workflows client. 
The two arguments can be seen here: https://github.com/dapr/durabletask-python/blob/main/durabletask/worker.py#L224-L225, and they are available since the `v0.2.0-alpha.8`, which is already being used.


nelson-parente (approved):

elena-kolevska (approved):

yevgen-el8:
The default value for `maximum_thread_pool_workers` defaults to `os.cpu_count() + 4`. It looks Python 3.12 doesn't respect container limits and on smaller nodes it might be a dangerously small value. I'd recommend to default it to the previous value of 16

PR #842, opened 2025-09-17
Adds support for interceptors and concurrency_options arguments in the workflow engine

acroca:
Brings https://github.com/dapr/python-sdk/pull/841 to `main`


acroca:
This PR is not necessary, we need to get `release-1.16` into `main` and #841 did it for 1.16.

PR #859, opened 2025-11-04
Put dev to 1.16.0.dev

acroca:
# Description

When we merged https://github.com/dapr/python-sdk/pull/851, we broke the dev builds.

`main` branch should remain being a `dev` build.


filintod:
I'll be setting a PR to remove Py3.9 as it is now out of its lifecycle

Issue #834, opened 2025-09-06
feat: asyncio support for workflow sdk


passuied:
## Describe the WORKFLOW SDK feature

The workflow sdks currently don't have asyncio versions like the DaprClient SDK.
This creates friction when using them within asyncio code as it leads to thread blocking/deadlock.
It would make sense to provide these, specifically:
- schedule_new_workflow()
- get_workflow_state()
- wait_for_workflow_start()
- wait_for_workflow_completion()
- raise_workflow_event()
- terminate_workflow()
- pause_workflow()
- resume_workflow()
- purge_workflow()

## Release Note

RELEASE NOTE: **ADD** asyncio versions of workflow sdk methods

passuied:
This has a dependency on async support in durabletask-python. I have created a [PR](https://github.com/microsoft/durabletask-python/pull/65/files#diff-84228e6734f8df7e909f5caa077a109ae829477939cee5156625b7487aac7344) to address it there first

PR #861, opened 2025-11-04
Added async workflow client implementation, leveraging new durabletask.aio.client implementation

passuied:
# Description

Added async workflow client implementation, leveraging new durabletask.aio.client implementation

## Issue reference

#834

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation - Pending Full async support for WorkflowRuntime

### DIFF
--- a/examples/demo_actor/README.md
+++ b/examples/demo_actor/README.md
@@ -17,15 +17,9 @@ This document describes how to create an Actor(DemoActor) and invoke its methods
 
 You can install dapr SDK package using pip command:
 
-<!-- STEP 
-name: Install requirements
--->
-
 ```sh
 pip3 install -r demo_actor/requirements.txt
 ```
-
-<!-- END_STEP -->
 
 ## Run in self-hosted mode
 
--- a/examples/demo_workflow/README.md
+++ b/examples/demo_workflow/README.md
@@ -18,15 +18,9 @@ It demonstrates the following APIs:
 
 You can install dapr SDK package using pip command:
 
-<!-- STEP
-name: Install requirements
--->
-
 ```sh
 pip3 install -r demo_workflow/requirements.txt
 ```
-
-<!-- END_STEP -->
 
 <!-- STEP
 name: Running this example


ext/dapr-ext-workflow/dapr/ext/workflow/aio/__init__.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+
+from .dapr_workflow_client import DaprWorkflowClientAsync
+
+# Public alias to mirror sync naming under aio namespace
+DaprWorkflowClient = DaprWorkflowClientAsync
+
+__all__ = [
+    'DaprWorkflowClientAsync',
+    'DaprWorkflowClient',
```
acroca: I think it'd be better to have only one exposed. I'd even rename the class to `DaprWorkflowClient` (dropping the `Async`) to match the `from dapr.aio.clients import DaprClient`. What do you think?

filintod:
@acroca actually this is a bit different, as this tackles the client initiating actions against a workflow(schedule workflow, signal events), instead of the internals of the workflow that my PR tackles. He did the PR in durabletask that complements this https://github.com/dapr/durabletask-python/pull/17.  The PRs his/mine are complementary and not overlapping

ext/dapr-ext-workflow/dapr/ext/workflow/aio/dapr_workflow_client.py
```diff
@@ -0,0 +1,280 @@
...
+from dapr.clients.http.client import DAPR_API_TOKEN_HEADER
+from dapr.conf import settings
+from dapr.conf.helpers import GrpcEndpoint
+
+T = TypeVar('T')
+TInput = TypeVar('TInput')
+TOutput = TypeVar('TOutput')
+
+
+class DaprWorkflowClientAsync:
```
acroca: I'd call this just `DaprWorkflowClient`, without the async, same as we do with `DaprClient` in both `dapr.aio.clients` and `dapr.clients`

ext/dapr-ext-workflow/dapr/ext/workflow/aio/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+]
```
acroca: Following the previous comment, I'd export only `DaprWorkflowClient`, since it's within `aio` it's going to be expected to be async.

acroca (commented):
Looks good, but I'd follow the same pattern as the `DaprClient` and not have a suffix `Async` for the async version.

acroca:
@passuied Can you add an example to `examples/workflow` that uses this new client? Maybe something like https://github.com/dapr/python-sdk/blob/main/examples/workflow/simple.py but async.
Sorry for adding new items to the review, but I didn't think about it when I reviewed yesterday 🤦

examples/workflow/simple_aio.py
```diff
@@ -0,0 +1,175 @@
+# -*- coding: utf-8 -*-
+# Copyright 2023 The Dapr Authors
```
acroca:
```suggestion
# Copyright 2025 The Dapr Authors
```

ext/dapr-ext-workflow/tests/aio/__init__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+
+"""
+Copyright 2023 The Dapr Authors
```
acroca:
```suggestion
Copyright 2025 The Dapr Authors
```

ext/dapr-ext-workflow/tests/aio/test_workflow_runtime.py
```diff
@@ -0,0 +1,176 @@
+# -*- coding: utf-8 -*-
+
+"""
+Copyright 2023 The Dapr Authors
```
acroca:
```suggestion
Copyright 2025 The Dapr Authors
```

examples/workflow/simple_aio.py
acroca:
```suggestion
# Copyright 2025 The Dapr Authors
```

ext/dapr-ext-workflow/dapr/ext/workflow/aio/workflow_runtime.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: utf-8 -*-
+
+"""
+Copyright 2023 The Dapr Authors
```
acroca:
```suggestion
Copyright 2025 The Dapr Authors
```

ext/dapr-ext-workflow/tests/aio/__init__.py
acroca:
```suggestion
Copyright 2025 The Dapr Authors
```

ext/dapr-ext-workflow/tests/aio/test_workflow_runtime.py
acroca:
```suggestion
Copyright 2025 The Dapr Authors
```

acroca (changes requested):
LGTM, but tests are failing, and copyright headers need to be fixed :)

ext/dapr-ext-workflow/setup.cfg
```diff
@@ -24,8 +24,8 @@ python_requires = >=3.9
 packages = find_namespace:
 include_package_data = True
 install_requires =
-    dapr >= 1.16.0.dev
-    durabletask-dapr >= 0.2.0a12
+    dapr >= 1.16.1rc1
```
acroca: We have to keep `1.16.0.dev` in the `main` branch

passuied: ok will revert

ext/dapr-ext-workflow/dapr/ext/workflow/aio/workflow_runtime.py
```diff
@@ -0,0 +1,149 @@
...
+                result = fn(*args, **kwargs)
+                if inspect.isawaitable(result):
+                    future = asyncio.run_coroutine_threadsafe(result, self._main_event_loop)
+                    return future.result()
+                return result
+        else:
+
+            @wraps(fn)
+            def target_fn():
+                return fn
```
akotsarelation: 1. Does this mean that if the `main_event_loop` was not provided in the constructor, we'll silently treat all activities as synchronous? If this is the case, should we make `main_event_loop` required? Otherwise, it seems to be defeating the purpose. Or am I missing something?
2. Isn't it supposed to call `fn()` instead of returning the function itself?

ext/dapr-ext-workflow/dapr/ext/workflow/aio/dapr_workflow_client.py
```diff
@@ -0,0 +1,280 @@
...
+        if hasattr(workflow, '_dapr_alternate_name'):
+            return await self.__obj.schedule_new_orchestration(
+                workflow.__dict__['_dapr_alternate_name'],
+                input=input,
+                instance_id=instance_id,
+                start_at=start_at,
+                reuse_id_policy=reuse_id_policy,
+            )
+        return await self.__obj.schedule_new_orchestration(
+            workflow.__name__,
```
akotsarelation: Minor - consider storing the `workflow.__dict__['_dapr_alternate_name']` / `workflow.__name__` in a variable and avoiding code duplication around calling `self.__obj.schedule_new_orchestration`

ext/dapr-ext-workflow/dapr/ext/workflow/aio/workflow_runtime.py
```diff
@@ -0,0 +1,149 @@
...
+
+ClientInterceptor = Union[
+    grpc.aio.UnaryUnaryClientInterceptor,
+    grpc.aio.UnaryStreamClientInterceptor,
+    grpc.aio.StreamUnaryClientInterceptor,
+    grpc.aio.StreamStreamClientInterceptor,
+]
+
+
+class WorkflowRuntime(WorkflowRuntimeSync):
```
akotsarelation: I'm not sure if wrapping the sync runtime is the best course of action here. Instead, it should be natively supported by `durabletask` it looks like there's been a lot of work, including async support done there. Are you sure it's not already a resolved problem?

passuied: @akotsarelation I agree that when I look at @filintod's [PR](https://github.com/dapr/durabletask-python/pull/13), I'm inclined to wait for his work to be complete so we can replace the whole `WorkflowRuntime` with his full async implementation...
It wasn't so apparent until I worked on the example @acroca and I had to override `WorkflowRuntime` (but not in an ideal way as I still need the work from that PR)... While the current solution works it does mix sync and async and comes with performance downsides... If this PR is coming soon, I'm inclined to not publish this one and instead wait for the full blown solution....

filintod: I guess the issue is that you were changing not only the client side (dapr_workflow_client) but also the workflowruntime, that yeah you would need something like what I am doing. Was that needed here? as your durabletask PR only dealt with the client side (schedule workflow, etc) and why I said this PR and the durabletask were complementary but not if mixing workflow runtime too.

filintod: I think that if you remove the `workflow_runtime` and the `test_workflow_runtime` files, then this PR would be fine. I was under the impression you wanted asyncio support for the client actions (schedule workflow, wait for workflow, raise events, etc).  Try in your example to use the regular sync WorkflowRuntime, but use your asyncio awaitable client methods to start the workflow and raise the events via await

passuied: Yes I only added it because I needed to provide an example and without a 'compatible' runtime, it didn't make sense at all...
If @acroca is ok I can revert workflowruntime, and example and we can bring back the example once we have all the pieces in place?

acroca: As long as everything in the PR is tested, I'm happy :P If I understand correctly, you'll remove the `test_workflow_runtime` but will keep the other right?

passuied: @acroca @filintod I removed all additional commits and only left aio.workflow_client and test_workflow_client.

ext/dapr-ext-workflow/dapr/ext/workflow/aio/workflow_runtime.py
passuied: removed this code

passuied:
Thanks @filintod that was helpful.

@acroca I have pushed a change with the simple.py example adapted for using the async client. Works very well!

examples/workflow/simple_aio_client.py
```diff
@@ -0,0 +1,173 @@
...
+        print('*** Workflow timed out!')
+
+    await wf_client.purge_workflow(instance_id=instance_id)
+    try:
+        await wf_client.get_workflow_state(instance_id=instance_id)
+    except DaprInternalError as err:
+        if non_existent_id_error in err._message:
+            print('Instance Successfully Purged')
+
+    wfr.shutdown()
```
filintod: add a try/finally to always execute `wfr.shudown()` on any issues, or the process might hang if any issue happens

passuied: done!

acroca:
```
❯ ruff format --check
Would reformat: examples/workflow/simple_aio_client.py
1 file would be reformatted, 229 files already formatted
```

passuied:
@acroca I should have addressed all outstanding issues... I found that some mm.py directive was overriding the local install of the ext packages...

examples/demo_workflow/README.md
```diff
@@ -18,16 +18,10 @@ It demonstrates the following APIs:
 
 You can install dapr SDK package using pip command:
 
-<!-- STEP
```
passuied, line 21: This part would get executed prior to the workflow example and would override the local package (installed in tox.ini as pre-command), leading to new module not found

examples/demo_actor/README.md
```diff
@@ -17,16 +17,10 @@ This document describes how to create an Actor(DemoActor) and invoke its methods
 
 You can install dapr SDK package using pip command:
 
-<!-- STEP
```
passuied, line 20: Applying the same fixes in `demo_actor` as I assume it would cause the same problem if the `dapr-ext-fastapi` server version overrides the local package installed by tox....

acroca (approved):

PR #868, opened 2025-11-22
Feat/ext langgraph checkpoint

CasperGN:
# Description

Extending `dapr-ext-langgraph` to expand on the handling for `put`, `put_writes` and `get_tuple` to conform with the native LangGraph `RedisSaver`. This ensures that the demo's presented in the [LangGraph Academy](https://github.com/langchain-ai/langchain-academy/tree/b20cf608f2bf165e09080961537d329f813cfb20) are functional.

## Issue reference

Please reference the issue this PR will close: N/A

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/examples/langgraph-checkpointer/README.md
+++ b/examples/langgraph-checkpointer/README.md
@@ -35,25 +35,13 @@ Run the following command in a terminal/command prompt:
 name: Run subscriber
 expected_stdout_lines:
   - '== APP == Add 3 and 4.'
-  - '== APP == Tool Calls:'
-  - '== APP ==   add'
-  - '== APP ==  Call'
   - '== APP ==     a: 3'
   - '== APP ==     b: 4'
-  - '== APP == Name: add'
   - '== APP == 7'
-  - '== APP == The result of adding 3 and 4 is 7.'
-  - '== APP == Multiply the result by 2.'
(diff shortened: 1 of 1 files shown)


acroca:
> @acroca do you want me to add the conditional run for the ext packages and ollama in this PR or should I rather add this in a follow up?
> 
> I'll also need to do a follow up to implement the delete function similar to the changes in this one

This PR adds roughly 2.5min to the CI build, right? I think it'd be good to split in a different runs, but I also want to experiment with running each validation in a different step to run.
Maybe let's merge as is, and we can improve CI runs in a separate PR.

CasperGN:
> > @acroca do you want me to add the conditional run for the ext packages and ollama in this PR or should I rather add this in a follow up?
> > I'll also need to do a follow up to implement the delete function similar to the changes in this one
> 
> This PR adds roughly 2.5min to the CI build, right? I think it'd be good to split in a different runs, but I also want to experiment with running each validation in a different step to run. Maybe let's merge as is, and we can improve CI runs in a separate PR.

Yeah and considering we'd be carrying more ext-<agent framework> packages in we should also have a conditional check on these so we don't have to pull ollama + llama3.2 every single time but only when the ext packages that needs it requires testing

PR #872, opened 2025-12-19
Add Strands Agent Session Manager

yaron2:
Adds the ability to store Strands Agents session state with Dapr.

### DIFF
--- a/ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
+++ b/ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
@@ -1,0 +1,536 @@
+"""Dapr state store session manager for distributed storage."""
+
+import json
+import logging
+from typing import Any, Dict, List, Literal, Optional, cast
+
+from dapr.clients import DaprClient
+from dapr.clients.grpc._state import Consistency, StateOptions
+
+from strands.types.exceptions import SessionException
+from strands.types.session import Session, SessionAgent, SessionMessage
+from strands.session.repository_session_manager import RepositorySessionManager
+from strands.session.session_repository import SessionRepository
+from strands import _identifier
+
+logger = logging.getLogger(__name__)
+
+# Type-safe consistency constants
+ConsistencyLevel = Literal["eventual", "strong"]
+DAPR_CONSISTENCY_EVENTUAL: ConsistencyLevel = "eventual"
+DAPR_CONSISTENCY_STRONG: ConsistencyLevel = "strong"
+
+class DaprSessionManager(RepositorySessionManager, SessionRepository):
+    """Dapr state store session manager for distributed storage.
+
+    Stores session data in Dapr state stores (Redis, PostgreSQL, MongoDB, Cosmos DB, etc.)
+    with support for TTL and consistency levels.
+
+    Key structure:
+    - `{session_id}:session` - Session metadata
+    - `{session_id}:agents:{agent_id}` - Agent metadata
+    - `{session_id}:messages:{agent_id}` - Message list (JSON array)
+    """
+
+    def __init__(
+        self,
+        session_id: str,
+        state_store_name: str,
+        dapr_client: DaprClient,
+        ttl: Optional[int] = None,
+        consistency: ConsistencyLevel = DAPR_CONSISTENCY_EVENTUAL,
+        **kwargs: Any,
+    ):
+        """Initialize DaprSessionManager.
+
+        Args:
+            session_id: ID for the session.
+                ID is not allowed to contain path separators (e.g., a/b).
+            state_store_name: Name of the Dapr state store component.
+            dapr_client: DaprClient instance for state operations.
+            ttl: Optional time-to-live in seconds for state items.
+            consistency: Consistency level for state operations ("eventual" or "strong").
+            **kwargs: Additional keyword arguments for future extensibility.
+        """
+        self._state_store_name = state_store_name
+        self._dapr_client = dapr_client
+        self._ttl = ttl
+        self._consistency = consistency
+        self._owns_client = False
+
+        super().__init__(session_id=session_id, session_repository=self)
+
+    @classmethod
+    def from_address(
+        cls,
+        session_id: str,
+        state_store_name: str,
+        dapr_address: str = "localhost:50001",
+        **kwargs: Any,
+    ) -> "DaprSessionManager":
+        """Create DaprSessionManager from Dapr address.
+
+        Args:
+            session_id: ID for the session.
+            state_store_name: Name of the Dapr state store component.
+            dapr_address: Dapr gRPC endpoint (default: localhost:50001).
+            **kwargs: Additional arguments passed to __init__ (ttl, consistency, etc.).
+
+        Returns:
+            DaprSessionManager instance with owned client.
+        """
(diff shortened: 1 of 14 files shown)


ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
```diff
@@ -0,0 +1,553 @@
...
+    """
+
+    def __init__(
+        self,
+        session_id: str,
+        state_store_name: str,
+        dapr_client: DaprClient,
+        ttl: Optional[int] = None,
+        consistency: ConsistencyLevel = DAPR_CONSISTENCY_EVENTUAL,
+        **kwargs: Any,
```
acroca: I'd remove this to avoid confusion, it's not being used

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
```diff
@@ -0,0 +1,553 @@
...
+                key=key,
+                options=self._get_state_options(),
+            )
+        except Exception as e:
+            raise SessionException(f'Failed to delete state key {key}: {e}') from e
+
+    def _get_manifest_key(self, session_id: str) -> str:
+        """Get session manifest key (tracks agent_ids for deletion)."""
+        session_id = _identifier.validate(session_id, _identifier.Identifier.SESSION)
+        return f'{session_id}:manifest'
```
acroca: Move this where the rest of the `_key` methods are please

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
```diff
@@ -0,0 +1,553 @@
...
+
+        # Check if session already exists
+        existing = self.read_session(session.session_id)
+        if existing is not None:
+            raise SessionException(f'Session {session.session_id} already exists')
+
+        # Write session data
+        session_dict = session.to_dict()
+        self._write_state(session_key, session_dict)
+        return session
```
acroca, line 274: Would this require some sort of locking mechanim? If we get two concurrent calls to `create_session`, both might end up writing the state and both will use the session but shouldn't.

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
```diff
@@ -0,0 +1,553 @@
...
+            content = json.dumps(data, ensure_ascii=False)
+            self._dapr_client.save_state(
+                store_name=self._state_store_name,
+                key=key,
+                value=content,
+                state_metadata=self._get_write_metadata(),
+                options=self._get_state_options(),
+            )
+        except Exception as e:
+            raise SessionException(f'Failed to write state key {key}: {e}') from e
```
acroca, line 232: Do you think it'd make sense to make use of etags to make sure we don't overwrite states?

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
```diff
@@ -0,0 +1,553 @@
...
+        Args:
+            session_id: ID of the session.
+            agent_id: ID of the agent.
+            session_message: Message to create.
+            **kwargs: Additional keyword arguments for future extensibility.
+
+        Raises:
+            SessionException: If creation fails.
+        """
+        messages_key = self._get_messages_key(session_id, agent_id)
```
acroca, line 405: Have you considered using a unique key per message? This would simplify read operations so we won't need to fetch all messages every time, but we'd need to keep track of all messages created somewhere so we can delete them all when deleting a session... What do you think?

ext/dapr-ext-strands/tests/test_session_manager.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
+
```
acroca, line 15: missing copyright header

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
```diff
@@ -0,0 +1,553 @@
+"""Dapr state store session manager for distributed storage."""
+
```
acroca: missing copyright header

ext/dapr-ext-strands/setup.cfg
```diff
@@ -0,0 +1,42 @@
+[metadata]
+url = https://dapr.io/
+author = Dapr Authors
+author_email = daprweb@microsoft.com
```
acroca, line 4: 👀

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
yaron2: All the existing implementations for Strands session managers (both in-tree and community contributed) assume the manager is not thread safe and do not coordinate locking. This is most likely because how Strands agents encapsulate sessions lifecycles and keeping access sequential

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
yaron2: We encountered the key/value data structure paradigm in both agents and non-agents workloads. In the end it's preferable to optimize for the least amount of I/O operations, especially when it comes to writing state. Not just to stop split/brain but to optimize for costs. Likely DynamoDB will be used a lot with Strands and we don't want to penalize two write operation for every message.

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
yaron2: I believe this would be tied to the locking question. I audited how the Strands built-in S3 file manager works, and it also does not use etags/record versioning even though it supports it. Again, I suspect this is because of the way Strands handles the access and lifecycle of a session manager in a way that makes this a non-issue.

ext/dapr-ext-strands/setup.cfg
yaron2: 🤷‍♂️

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
yaron2: Done

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
yaron2: Agree, removed

ext/dapr-ext-strands/tests/test_session_manager.py
yaron2: Added

ext/dapr-ext-strands/dapr/ext/strands/dapr_session_manager.py
yaron2: Added

yaron2:
@acroca All feedback and questions addressed. Examples seem to be failing regardless.

acroca (approved):

marcduiker:
@holopin-bot @yaron2 sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #886, opened 2026-01-14
Trigger ci for new ext tags

acroca:
We've recently added new `ext` packages but didn't add them to the ci triggers.


CasperGN (approved):
LGTM - @acroca I'll put it on my mental backlog to come back and make the trigger for tests for strands/langgraph to be dependent on changes in the folder structure. @sicoyle and I are also working on having more stable llm runs in ci and will come back with these updates once ready.

PR #901, opened 2026-01-29
fix: signal when dt reader stream is ready within wf client start call

sicoyle:
# Description

Corresponding PR for https://github.com/dapr/durabletask-python/pull/42

I identified a race condition where we report success from wf_client.start() in the Python SDK before the DurableTask worker’s stream reader thread is fully initialized. This PR addresses the issue by introducing a poll to check for the worker and its stream reader thread have actually started so it can actually signal everything is ready and the wf client start was fully successful. This will ensure that workflows scheduled immediately do not hit issues due to the stream reader not being ready yet. I also added more try/except blocks and additional logging.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -174,13 +194,80 @@ def activityWrapper(ctx: task.ActivityContext, inp: Optional[TInput] = None):
...
+            except Exception as start_error:
+                self._logger.exception(
+                    f"WorkflowRuntime worker did not start: {start_error}"
+                )
+                raise
+
+            # Verify the worker and its stream reader are ready
+            if hasattr(self.__worker, 'is_worker_ready'):
+                try:
+                    is_ready = self.wait_for_worker_ready(timeout=5.0)
```
CasperGN: Should the timeout not be set-able? Otherwise what's the point of having the default of `30.0`?

CasperGN (changes requested):

CasperGN (dismissed):

PR #916, opened 2026-02-12
Added support for python 3.14 and done some pending 3.9 removals

acroca:
- Added support for 3.14
- Some chores removing 3.9 from missing places.


sicoyle (approved):
thank you!!

PR #942, opened 2026-03-06
feat: Add aio module and update mypy configuration to include aio files

acroca:
This change should fix the mypy issue where we have to ignore some checks when importing the `aio` client ``.


sicoyle (approved):
thank you!!

PR #956, opened 2026-03-12
No 60s timeout by default

cicoyle:
Remove hardcoded 60s default timeout. Only apply if explicitly configured. 

I was running a long running agent/wf and it was being killed by this and it was not respecting the resiliency configured. An SDK-level default deadline that fires before resiliency can act prevents retries from ever being attempted...


tests/clients/test_timeout_interceptor.py
```diff
@@ -54,3 +54,20 @@ def test_intercept_unary_unary_without_timeout(self):
         )
         called_client_call_details = continuation.call_args[0][0]
         self.assertEqual(7, called_client_call_details.timeout)
+
+    @patch.object(settings, 'DAPR_API_TIMEOUT_SECONDS', None)
+    def test_intercept_unary_unary_no_global_timeout_no_per_call_timeout(self):
```
acroca, line 79: Don't we need a test that covers `DAPR_API_TIMEOUT_SECONDS` being `None` and `timeout` having a value?

sicoyle (approved):
thank you!

PR #958, opened 2026-03-13
fix(build): handle rate limiting causing failures in build

sicoyle:
# Description

There have been several PRs such as this one with 3.14 failing the build on the validation of the quickstarts: https://github.com/dapr/python-sdk/actions/runs/23053657978/job/66961073927

This is because curl fetches the latest version and gets rate-limited since this is unathenticated calls, and then DAPR_CLI_VER is an empty string, which yields the broken zip URL of `download//dapr_linux_amd64.tar.gz`.

This PR makes the curl authenticated which should increase our rate limits and use proper json parsing with jq and more reliable checks for proper installation.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/validate_examples.yaml
+++ b/.github/workflows/validate_examples.yaml
@@ -99,7 +99,7 @@ jobs:
           pip install setuptools wheel twine tox
       - name: Set up Dapr CLI
         run: |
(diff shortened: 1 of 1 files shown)


acroca (approved):
@sicoyle Soon we'll be able to move to use shared workflow actions, see: https://github.com/dapr/.github/pull/8

PR #960, opened 2026-03-13
Migrate to uv

acroca:
## Summary

- Replace `tox.ini`, `setup.cfg`, `mypy.ini`, and `dev-requirements.txt` with `pyproject.toml` (PEP 621) and a `uv` workspace
- Add per-extension `pyproject.toml` files with `[tool.uv.sources]` workspace references
- Update all CI workflows to use `uv sync --frozen` / `uv run` for lint, typecheck, tests, and example validation
- Migrate tag-based publish steps from `python setup.py sdist bdist_wheel` to PEP 517 `python -m build`
- Fix `dapr-dev` dependency rewrite using `packaging.requirements.Requirement` instead of fragile `startswith` check
- Fix shell quoting bugs in CI autoformatter steps (`$statusResult` → `"$statusResult"`)
- Add input validation and proper quoting to `examples/validate.sh`


.github/workflows/build-push-to-main.yaml
```diff
@@ -44,16 +44,23 @@ jobs:
...
       run: |
-        tox -e py`echo "${{ matrix.python_ver }}" | sed 's/\.//g'`
+        uv run coverage run -m unittest discover -v ./tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-workflow/tests
+        uv run coverage run -a -m pytest -m "not e2e" ./ext/dapr-ext-workflow/tests/durabletask/
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-grpc/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-fastapi/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-langgraph/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-strands/tests
+        uv run coverage run -a -m unittest discover -v ./ext/flask_dapr/tests
```
sicoyle: can we specify the dir ./ext/ instead of every subdir here?

.github/workflows/build-tag.yaml
```diff
@@ -44,16 +44,23 @@ jobs:
...
-        tox -e py`echo "${{ matrix.python_ver }}" | sed 's/\.//g'`
+        uv run coverage run -m unittest discover -v ./tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-workflow/tests
+        uv run coverage run -a -m pytest -m "not e2e" ./ext/dapr-ext-workflow/tests/durabletask/
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-grpc/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-fastapi/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-langgraph/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-strands/tests
+        uv run coverage run -a -m unittest discover -v ./ext/flask_dapr/tests
+        uv run coverage xml
```
sicoyle, line 60: same comment here on can we put the ./ext dir instead of each subdir here?

.github/workflows/build.yaml
```diff
@@ -50,15 +50,22 @@ jobs:
...
       run: |
-        tox -e py`echo "${{ matrix.python_ver }}" | sed 's/\.//g'`
+        uv run coverage run -m unittest discover -v ./tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-workflow/tests
+        uv run coverage run -a -m pytest -m "not e2e" ./ext/dapr-ext-workflow/tests/durabletask/
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-grpc/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-fastapi/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-langgraph/tests
+        uv run coverage run -a -m unittest discover -v ./ext/dapr-ext-strands/tests
+        uv run coverage run -a -m unittest discover -v ./ext/flask_dapr/tests
```
sicoyle: is there a way to not have to specify each subdir here?

.github/workflows/validate_examples.yaml
```diff
@@ -153,4 +153,26 @@ jobs:
...
+          ./validate.sh demo_actor
+          ./validate.sh invoke-binding
+          ./validate.sh grpc_proxying
+          ./validate.sh w3c-tracing
+          ./validate.sh distributed_lock
+          ./validate.sh configuration
+          ./validate.sh workflow
+          ./validate.sh jobs
+          ./validate.sh langgraph-checkpointer
+          ./validate.sh ../
```
sicoyle, line 163: i feel like this might make it easy to get out of sync and miss some... is there a way to use the examples dir instead and point to it?

ext/dapr-ext-fastapi/setup.py
```diff
@@ -13,9 +13,15 @@
 limitations under the License.
 """
 
-import configparser
 import os
+import sys
 
+if sys.version_info >= (3, 11):
```
sicoyle, line 19: i see python 3.10 supported elsewhere... so pls add 3.10 here

ext/dapr-ext-grpc/pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "dapr>=1.17.0.dev0",
```
sicoyle: should we use a non dev version here

ext/dapr-ext-grpc/pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+    "cloudevents>=1.0.0,<2.0.0",
+]
+
+[project.urls]
+Documentation = "https://github.com/dapr/docs"
+Source = "https://github.com/dapr/python-sdk"
+
+[build-system]
+requires = ["setuptools>=61.0", "wheel", "packaging", "tomli; python_version < '3.11'"]
```
sicoyle, line 31: 3.10

ext/dapr-ext-grpc/setup.py
```diff
@@ -13,9 +13,15 @@
 limitations under the License.
 """
 
-import configparser
 import os
+import sys
 
+if sys.version_info >= (3, 11):
```
sicoyle, line 19: 3.10 pls

sicoyle (changes requested):
few comments so far - thank you bc this will be a great improvement 🙌

ext/dapr-ext-fastapi/setup.py
acroca: 3.10 is supported in the `else`. For 3.11+ we use the stdlib, but for 3.10 we use `tomli`

.github/workflows/validate_examples.yaml
acroca: Replaced with a loop

ext/dapr-ext-fastapi/pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "dapr>=1.17.0",
```
sicoyle, line 22:
```suggestion
    "dapr>=1.17.4",
```

sicoyle: https://pypi.org/project/dapr/

ext/dapr-ext-grpc/pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "dapr>=1.17.0",
```
sicoyle, line 22:
```suggestion
    "dapr>=1.17.4",
```

ext/dapr-ext-grpc/pyproject.toml
sicoyle: is this resolved? We support down to 3.10... so why is this 3.11?

ext/dapr-ext-workflow/setup.py
```diff
@@ -13,9 +13,15 @@
 limitations under the License.
 """
 
-import configparser
 import os
+import sys
 
+if sys.version_info >= (3, 11):
```
sicoyle, line 19: isnt this supposed to be 3.10 as lowest?

ext/dapr-ext-grpc/pyproject.toml
acroca: `tomli` is only added as an external dependency in 3.10 because it was added to the stdlib in 3.11. That's why the dependency has this condition, as well as the imports.

ext/dapr-ext-fastapi/pyproject.toml
acroca: I think `main` will point to `1.17.0`, but the 'real' dependency is added in the `release-1.17` branch, right?

sicoyle: Yes, I think main should use the latest available version. Even within the same minor version, 1.17.4 over 1.17.0 likely includes bug fixes and improvements we'd want to build against on main. We follow the same pattern with other dependencies like durabletask, kit, etc. — bumping to the latest patch on main before an official release. The "real" pinned dependency can live on the release branch, but there's no reason to leave main lagging behind on a known good/newer patch version.

ext/flask_dapr/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "Flask>=1.1",
```
sicoyle: 1.1 is super low. It also does not follow semantic conventions. There is a flask 3.1.3 can you bump us or at least use a major.minor.patch version?

pyproject.toml
```diff
@@ -1,3 +1,88 @@
...
+
+[dependency-groups]
+dev = [
+    "mypy>=1.2.0",
+    "mypy-extensions>=0.4.3",
+    "mypy-protobuf>=2.9",
+    "types-python-dateutil",
+    "types-grpcio-status",
+    "coverage>=5.3",
+    "wheel",
```
sicoyle: pls pin a version here

pyproject.toml
```diff
@@ -1,3 +1,88 @@
...
+]
+
+[dependency-groups]
+dev = [
+    "mypy>=1.2.0",
+    "mypy-extensions>=0.4.3",
+    "mypy-protobuf>=2.9",
+    "types-python-dateutil",
+    "types-grpcio-status",
+    "coverage>=5.3",
```
sicoyle: pls pin to a major.minor.patch version for proper semantic conventions

pyproject.toml
```diff
@@ -1,3 +1,88 @@
...
+    "wheel",
+    "opentelemetry-sdk",
+    "opentelemetry-instrumentation-grpc",
+    "opentelemetry-exporter-zipkin",
+    "httpx>=0.24",
+    "pyOpenSSL>=23.2.0",
+    "Flask>=1.1",
+    "pytest>=7.0",
```
sicoyle: major.minor.patch versions pls

sicoyle (commented):
few more commetns pls

ext/flask_dapr/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+    "dapr>=1.17.0",
```
seherv, line 23: Should we bump this to 1.17.4 too? @sicoyle

AGENTS.md
```diff
@@ -184,14 +169,14 @@ When completing any task on this project, work through this checklist. Not every
...
-- [ ] Run `mypy --config-file mypy.ini` if you changed files covered by mypy (actor, clients, conf, serializers, ext-grpc, ext-fastapi, flask_dapr)
+- [ ] Run `uv run ruff check --fix && uv run ruff format` and fix any remaining issues
+- [ ] Run `uv run mypy` if you changed files covered by mypy (actor, aio, clients, conf, serializers, ext-grpc, ext-fastapi, flask_dapr)
 
 ### Examples (integration tests)
 
 - [ ] If you added a new user-facing feature or building block, add or update an example in `examples/`
 - [ ] Ensure the example README has `<!-- STEP -->` blocks with `expected_stdout_lines` so it is validated in CI
-- [ ] If you added a new example, register it in `tox.ini` under `[testenv:examples]`
+- [ ] If you added a new example, add it to the `validate_examples.yaml` CI workflow
```
seherv: I believe the loop over the examples/ folder [you added recently](https://github.com/dapr/python-sdk/pull/960/changes#r3051227442) makes this unnecessary

ext/dapr-ext-fastapi/pyproject.toml
acroca: Maybe related to the discussions here: https://github.com/dapr/python-sdk/issues/961, but shouldn't this dependency in `main` be something like this?:
```
dapr@git+https://github.com/dapr/python-sdk.git@main
```
This way, if someone uses the ext package from the main branch, they will also get the main `dapr` package from the main branch.

acroca: I implemented this change. With the dependency pointing to git, when a user fetches our ext packages from git they will also get the `dapr` dependency from git, which is the expectation.
For local development of the SDK we have `[tool.uv.sources]`, which basically tells uv to use the local workspace for the `dapr` package, so it won't fetch from git in this case.
With this, I think #961 should be closed.

ext/dapr-ext-workflow/setup.py
acroca: As mentioned in a separated related thread, 3.10 needs `tomli` as a dependency, but from 3.11 the stdlib includes it, so doesn't require the external dependency, and the import is different.

ext/flask_dapr/pyproject.toml
acroca: I would like to keep existing dependencies in this PR to keep it simpler

ext/flask_dapr/pyproject.toml
acroca: As mentioned previously, I changed the dependency to `dapr` to use the github reference. This reference would live only in `main`, the numbered release will need to point to the right number.

ext/dapr-ext-fastapi/pyproject.toml
acroca: And now I reverted it back to 1.7.0. I think we should address #961 in a different PR as it affects many other things, not just this version dependency.
I went back to 1.7.0 because this is the version we've always used for dev releases: https://pypi.org/project/dapr-dev/#history
In dev packages we always push to `-dev`, so I don't think patch version really matters. And in release branches, we always pin it to the right patch version, so I think we should keep it at `.0` in `main`.

CasperGN: I don't think this is the correct approach @acroca. 
Dapr is in-repo. These packages should be workspace packages and then `dapr` should be referred to as 
```
dependencies = [
    "dapr",
    ...
]
[tool.uv.sources]
dapr = { workspace = true }
```

This is exactly the kind of thing `uv` does well and not addressing it in this PR misses a big part of the point of even doing this migration to begin with.

ext/dapr-ext-langgraph/pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "dapr>=1.17.0",
```
CasperGN:
```suggestion
    "dapr",
```

ext/dapr-ext-grpc/pyproject.toml
CasperGN:
```suggestion
    "dapr",
```

ext/dapr-ext-fastapi/pyproject.toml
CasperGN:
```suggestion
    "dapr",
```

ext/dapr-ext-strands/pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "dapr>=1.17.0",
```
CasperGN:
```suggestion
    "dapr",
```

ext/dapr-ext-workflow/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "dapr>=1.17.0",
```
CasperGN:
```suggestion
    "dapr",
```

ext/flask_dapr/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "Flask>=1.1.4,<4.0.0",
+    "dapr>=1.17.0",
```
CasperGN:
```suggestion
    "dapr",
```

CasperGN (changes requested):
Don't specify the version when serving the dependency from a workspace member

ext/dapr-ext-fastapi/pyproject.toml
acroca: @CasperGN if I'm not mistaken the `tool.uv.sources` (which is added in this PR) applies only when doing sdk work (and CI), but if someone points their `dapr-ext-workflow` to github instead of pypi, they will get the `dapr` dependency indicated in the `dependencies`.
So when we publish the `-dev` package to [pypi](https://pypi.org/project/dapr-dev/#history), so far we've been pushing versions like `1.17.0.dev103`, even if the `release-1.17` is in 1.17.3. In this PR I didn't change any of that, and since we're planning to work on getting rid of `-dev` in #961, I thought it was better to keep it working as today. So the version indicated in the `dependencies` is really only used if a dependency is pointed to the github repo, because the pypi `-dev` package is manipulated in the `setup.py` to point to `dapr-dev` too...

ext/dapr-ext-grpc/pyproject.toml
acroca: Removed the version as suggested by Casper

ext/dapr-ext-fastapi/pyproject.toml
acroca: Removed the version, having it to support dependencies pointing through git is not worth the complexity of keeping it.

sicoyle (dismissed):
thank you!

CasperGN (approved):

PR #963, opened 2026-03-24
refactor: mv durabletask-python into sdk

sicoyle:
# Description

Move durabletask python into python sdk
https://github.com/dapr/durabletask-python

1. Moved durabletask/ into ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/ — vendored as a private internal package, no more separate PyPI dependency
2. Rewrote all imports across ext-workflow source, tests, and examples from durabletask.* to dapr.ext.workflow._durabletask.*; removed the try/except import fallbacks
3. Updated CI — deleted durabletask-validation.yml, removed standalone publish/test steps from build workflows; durabletask now tested and linted alongside everything else
4. Updated license headers on all vendored files from Microsoft/MIT to Dapr/Apache 2.0 (2026)
5. Cleaned up build config — removed durabletask-dapr from setup.cfg, dev-requirements.txt, and tox.ini; added pytest for durabletask unit tests; narrowed ruff exclude to only proto files

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ext/dapr-ext-workflow/dapr/ext/workflow/logger/logger.py
```diff
@@ -13,7 +13,8 @@ def __init__(self, name: str, options: Union[LoggerOptions, None] = None):
         log_handler.setLevel(options.log_level)
         log_handler.setFormatter(options.log_formatter)
         logger = logging.getLogger(name)
-        logger.handlers.append(log_handler)
+        if not logger.handlers:
+            logger.addHandler(log_handler)
```
acroca, line 17: Do we need the check?
```suggestion
        logger.addHandler(log_handler)
```

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
```diff
@@ -18,18 +18,21 @@
...
 from dapr.ext.workflow.util import getAddress
 from dapr.ext.workflow.workflow_context import Workflow
 from dapr.ext.workflow.workflow_state import WorkflowState
-from durabletask import client
 from grpc import RpcError
 
+try:
+    import durabletask.internal.protos as pb
+except ImportError:
+    import durabletask.internal.orchestrator_service_pb2 as pb  # type: ignore[no-redef]
```
acroca: Why is this fallback necessary?

tox.ini
```diff
@@ -94,8 +95,9 @@ commands =
 allowlist_externals=*
 
 commands_pre =
-    pip uninstall -y dapr dapr-ext-grpc dapr-ext-fastapi dapr-ext-langgraph dapr-ext-strands dapr-ext-flask dapr-ext-langgraph dapr-ext-strands
-    pip install -e {toxinidir}/ \
+    pip uninstall -y dapr dapr-ext-grpc dapr-ext-fastapi dapr-ext-langgraph dapr-ext-strands dapr-ext-flask dapr-ext-langgraph dapr-ext-strands durabletask-dapr
+    pip install -e {toxinidir}/durabletask/ \
```
acroca: Why don't we put durabletask in `dapr.durabletask`? Or even in the `ext-workflow` if it's only needed there. I don't think it needs to have its own package

.github/workflows/durabletask-validation.yml
```diff
@@ -0,0 +1,73 @@
...
+      with:
+        go-version: 'stable'
+    # Install and run the durabletask-go sidecar for running e2e tests
+    - name: Pytest e2e tests
+      run: |
+        # TODO: use dapr run instead of durabletask-go as it provides a more reliable sidecar behavior for e2e tests
+        go install github.com/dapr/durabletask-go@main
+        durabletask-go --port 4001 &
+        pytest -m e2e durabletask_tests/
+    - name: Run examples
```
acroca: I think this should happen as part of the `validate_examples` step

.github/workflows/durabletask-validation.yml
```diff
@@ -0,0 +1,73 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: DurableTask Python Build Validation
```
acroca: Shouldn't durabletask be tested in the same steps as everything else?

acroca (changes requested):
I think we need to change all license headers to use dapr's license right?

ext/dapr-ext-workflow/dapr/ext/workflow/logger/logger.py
sicoyle: yes, this was a copilot feedback fix as well as I noticed double logging here and there personally that this addresses

ext/dapr-ext-workflow/dapr/ext/workflow/dapr_workflow_client.py
sicoyle: no longer needed with my last refactor

acroca (commented):
LGTM, but we should add a way to build the protos for durabletask. Maybe a script in `./tools` like the existing one? In durabletask we have a make target, but since we don't have Make in the sdk I think the script would be more consistent.

acroca (approved):

PR #970, opened 2026-03-25
Use single `v*` tags to release

### LINKED ISSUE #969: Simplify release process
At the moment, when releasing a new version we need to push all the tags for the main `dapr` package and the ext packages. This is inconvenient for maintainers, as we have to do it manually every time and there's a risk of mistake.

I suggest we go with a single tag like `v1.17.2`, and that single tag would release everything at once. With this in place, we can even use the standard github release process, where we create a Release and it creates a tag automatically.

acroca:
Close https://github.com/dapr/python-sdk/issues/969

Simplifies release process by using only a single `v*` tag instead of a tag per ext package.

With this, we can trigger releases using github releases.

### DIFF
--- a/.github/workflows/build-tag.yaml
+++ b/.github/workflows/build-tag.yaml
@@ -4,12 +4,6 @@ on:
   push:
     tags:
       - v*
-      - workflow-v*
-      - grpc-v*
-      - flask-v*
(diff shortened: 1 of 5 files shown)


sicoyle (approved):
thank youuu!

PR #976, opened 2026-04-10
Populate timer origin field

acroca:
- Rebuilt latest protos
- Populate the new `origin` field in timers.
- Introduce automatic timer for `wait_for_external_event` to create a timer associated with such event. Other SDKs follow this same approach.


ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/helpers.py
```diff
@@ -11,11 +11,73 @@
 
 import traceback
 from datetime import datetime
-from typing import Optional
+from typing import Optional, Union
 
 import dapr.ext.workflow._durabletask.internal.protos as pb
 from google.protobuf import timestamp_pb2, wrappers_pb2
 
+TimerOrigin = Union[
```
sicoyle: can you mv these changes into a timer.py file instead of adding to the helpers.py pls

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/helpers.py
```diff
@@ -11,11 +11,73 @@
...
+        return False
+    timer = event.timerCreated
+    if (
+        timer.fireAt.seconds != OPTIONAL_TIMER_FIRE_AT.seconds
+        or timer.fireAt.nanos != OPTIONAL_TIMER_FIRE_AT.nanos
+    ):
+        return False
+    origin = timer.WhichOneof('origin')
+    return origin in (None, 'externalEvent')
+
```
sicoyle: these look almost like the same func... can we create a helper for the same logic between the two?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/helpers.py
```diff
@@ -11,11 +11,73 @@
...
+    uses this check to drop the optional action and shift sequence ids.
+    """
+    if not action.HasField('createTimer'):
+        return False
+    timer = action.createTimer
+    if timer.WhichOneof('origin') != 'externalEvent':
+        return False
+    return (
+        timer.fireAt.seconds == OPTIONAL_TIMER_FIRE_AT.seconds
+        and timer.fireAt.nanos == OPTIONAL_TIMER_FIRE_AT.nanos
```
sicoyle: will both always be set?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/helpers.py
```diff
@@ -11,11 +11,73 @@
...
+    fireAt as optional even if the proto3 ``origin`` oneof is unset (e.g. when
+    reading histories emitted by older sidecars that didn't populate it). When
+    ``origin`` *is* populated, it must match TimerOriginExternalEvent.
+    """
+    if not event.HasField('timerCreated'):
+        return False
+    timer = event.timerCreated
+    if (
+        timer.fireAt.seconds != OPTIONAL_TIMER_FIRE_AT.seconds
+        or timer.fireAt.nanos != OPTIONAL_TIMER_FIRE_AT.nanos
```
sicoyle: why is this one or if the func above uses and for this check?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/helpers.py
```diff
@@ -202,10 +272,20 @@ def new_workflow_version_not_available_action(
...
-    timestamp = timestamp_pb2.Timestamp()
-    timestamp.FromDatetime(fire_at)
-    return pb.WorkflowAction(id=id, createTimer=pb.CreateTimerAction(fireAt=timestamp))
+def new_create_timer_action(
+    id: int,
+    fire_at: Union[datetime, timestamp_pb2.Timestamp],
+    origin: Optional[TimerOrigin] = None,
+) -> pb.WorkflowAction:
+    if isinstance(fire_at, timestamp_pb2.Timestamp):
+        timestamp = fire_at
```
sicoyle: consistency between timer_id and id params between the two funcs and also vars within on the ts vs timestamp for the same thing pls

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/history_events_pb2.pyi
```diff
@@ -351,6 +351,44 @@ class TimerOriginExternalEvent(_message.Message):
 
 Global___TimerOriginExternalEvent: _TypeAlias = TimerOriginExternalEvent  # noqa: Y015
 
+@_typing.final
+class TimerOriginActivityRetry(_message.Message):
+    """Indicates the timer was created as a retry delay for an activity execution."""
+
+    DESCRIPTOR: _descriptor.Descriptor
+
+    TASKEXECUTIONID_FIELD_NUMBER: _builtins.int
```
sicoyle, line 360:
```suggestion
    TASK_EXECUTION_ID_FIELD_NUMBER: _builtins.int
```

why do we need to track the field number for this? can it just be TASK_EXECUTION_ID instead?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/task.py
```diff
@@ -166,21 +166,33 @@ def call_sub_orchestrator(
...
+            Controls how long to wait for the event. Three shapes:
+
+            * ``None`` (default) or a *negative* ``timedelta`` — wait indefinitely.
+              An optional sentinel timer is scheduled internally for runtime
+              tracking, but ``TimeoutError`` is never raised on its own.
+            * ``timedelta(0)`` — do not wait at all. The returned task fails
+              immediately with ``TimeoutError``.
+            * A future ``datetime`` or a positive ``timedelta`` — wait until
+              that deadline / for that duration; ``TimeoutError`` is raised if
+              the event has not been received in time.
```
sicoyle, line 191: which of these does plain 0 fall under? How do these compare to the other SDK timeout field defaults? How does this compare to other workflow providers for default behaviors?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/task.py
```diff
@@ -397,6 +409,19 @@ def fail(self, message: str, details: pb.TaskFailureDetails):
...
             self._parent.on_child_completed(self)
 
+    def cancel(self, exc: Exception) -> None:
+        """Mark this task as completed with the given exception.
+
+        Used for non-runtime failures such as a zero-timeout
+        wait_for_external_event being canceled before any history event arrives.
+        """
+        if self._is_complete:
+            raise ValueError('The task has already completed.')
```
sicoyle: should this instead be a noop?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/task.py
```diff
@@ -494,6 +519,58 @@ def on_child_completed(self, task: Task):
...
+        self._event_name = event_name
+        self._timeout = timeout
+        self._on_timeout = on_timeout
+        super().__init__([event_task, timer_task])
+
+    def on_child_completed(self, completed_task: Task):
+        if self.is_complete:
+            return
+        if completed_task is self._event_task:
+            if completed_task.is_failed:
```
sicoyle, line 562: this is missing a few other terminal states i believe in the if conditional check

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/task.py
```diff
@@ -494,6 +519,58 @@ def on_child_completed(self, task: Task):
...
+        elif completed_task is self._timer_task:
+            if self._on_timeout is not None:
+                self._on_timeout()
+            if self._timeout is not None:
+                msg = (
+                    f'Timed out after {self._timeout!r} waiting for '
+                    f'external event {self._event_name!r}'
+                )
+            else:
+                msg = f'Timed out waiting for external event {self._event_name!r}'
```
sicoyle: can these just be combined instead of split in an if else?

sicoyle (changes requested):
nice job 👏 👏 👏 🙌 few comments so far and I've reviewed mostttt files

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/helpers.py
acroca: Yes, updated the docstring to be clear.

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/task.py
acroca: `on_child_completed` is only called after a child completes, so it's guaranteed to be in one of two states

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/history_events_pb2.pyi
acroca: This is generated from the protos: https://github.com/dapr/durabletask-protobuf/blob/3380843f88b6df933600d16502725fe99591bdee/protos/history_events.proto#L96

sicoyle (approved):
LGTM - thank you 👏

PR #978, opened 2026-04-14
Fix when-all buggy error handling

acroca:
when using a `when_all` and one of the tasks fail, the resulting task sometimes failed, sometimes didn't.


sicoyle (approved):
thank you!!

PR #979, opened 2026-04-14
Fix CI build

acroca:
We're having some build failure sin `main`: https://github.com/dapr/python-sdk/pull/978


sicoyle (approved):
thank you!!

PR #982, opened 2026-04-17
Support Pydantic models

### LINKED ISSUE #858: [WORKFLOW SDK FEATURE REQUEST] Support pydantic models
It would be convenient if workflows/activities could support Pydantic models as input, automatically handling the (de)serializaiton. Right now, they only support dictionaries or simple values. The motivation comes from wanting to use strongly typed objects vs dictionaries (which are more opaque in their types).

There is some precedent here with [the FastAPI extension supporting them in subscriptions](https://docs.dapr.io/developing-applications/sdks/python/python-sdk-extensions/python-fastapi/).

## Release Note

RELEASE NOTE: Workflows and activities can now use Pydantic models as their input.

acroca:
## Summary
Closes https://github.com/dapr/python-sdk/issues/858

Users can now pass Pydantic `BaseModel` instances directly to `schedule_new_workflow`, `call_activity`, `call_child_workflow`, and `raise_workflow_event`, and receive real model instances on the other side just by annotating their workflow/activity input parameter — no manual `model_dump` / `model_validate` calls required.

```python
from pydantic import BaseModel
from dapr.ext.workflow import WorkflowRuntime, DaprWorkflowContext, WorkflowActivityContext

class OrderRequest(BaseModel):
    order_id: str
    amount: float

wfr = WorkflowRuntime()

@wfr.workflow
def order_workflow(ctx: DaprWorkflowContext, order: OrderRequest):
    # `order` is a real OrderRequest instance, not a dict
    yield ctx.call_activity(charge, input=order)

@wfr.activity
def charge(ctx: WorkflowActivityContext, order: OrderRequest):
    # Same here — the runtime reads the annotation and reconstructs the model
    return {'charged': order.amount}
```

## Design

**Encoding (send side):** `InternalJSONEncoder.default()` gets a new branch that detects objects implementing the Pydantic v2 model protocol (`model_dump` + `model_validate`) and emits a plain JSON object — no `AUTO_SERIALIZED` marker. The wire format stays interop-friendly with non-Python SDKs, and cross-app receivers that don't import the same class still see a readable payload.

**Decoding (receive side):** `WorkflowRuntime.register_workflow`, `register_versioned_workflow`, and `register_activity` inspect the user function's signature once at registration time. If the input parameter is annotated with a model-protocol class (including `Optional[Model]` / `Model | None`), the wrapper auto-coerces the incoming dict into a real instance before calling the user's function. No annotation → unchanged behavior (today's `SimpleNamespace` / dict path).

**Why duck-typing on `model_dump` / `model_validate`, not `isinstance(BaseModel)`:**
- **No hard dependency on pydantic** — the SDK never imports it. Users who don't use pydantic are unaffected.
- **Works with anything mirroring the protocol** — Pydantic v2, SQLModel, custom classes, future libraries.
- **Pydantic v1 is naturally excluded** — v1 uses `.dict()` / `.parse_obj()`, not `model_dump` / `model_validate`, so no version-check gymnastics.

`dump_model` inspects the target class's `model_dump` signature once (cached with `lru_cache`) and passes `mode='json'` when supported, falling back to bare `model_dump()` otherwise. Real Pydantic models always get `mode='json'` so nested datetimes / enums / UUIDs render as JSON-safe primitives.

**What this PR deliberately does NOT do:**
- No wire-format marker or FQN-based class lookup. Security and cross-language portability win out over full automatic round-tripping of yielded results.
- No `return_type=` kwarg on `call_activity` / `call_child_workflow`. Results arriving through `yield` come back as dicts; users write `Model.model_validate(result)` in one line when they want a typed instance. This avoids a new public API surface and a non-trivial task-proxy implementation in the workflow context.
- No changes to the vendored `_durabletask` worker machinery beyond the single encoder branch. Future upstream pulls stay clean.

## Changes

**Source (3 files):**
- `ext/dapr-ext-workflow/dapr/ext/workflow/_model_protocol.py` — new. Duck-typed helpers: `is_model`, `is_model_class`, `dump_model`, `coerce_to_model`, `resolve_input_model`. ~130 lines.
- `ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/shared.py` — one `if` branch in `InternalJSONEncoder.default()`, clearly marked `# Dapr-specific`. ~10 lines.
- `ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py` — resolve the input model once per registration and coerce in each wrapper. ~9 lines across three sites.

**Tests (3 files, all Pydantic-primary):**
- `ext/dapr-ext-workflow/tests/test_model_protocol.py` — new. Model-protocol helpers exercised with real Pydantic models; two small duck-typed tests cover the `mode='json'`-fallback path and partial-implementation rejection (behaviors Pydantic v2 can't exercise).
- `ext/dapr-ext-workflow/tests/test_workflow_runtime.py` — `PydanticInputCoercionTest` covering activity + workflow + versioned-workflow paths, `Optional[Model]`, passthrough of existing instances, pydantic `ValidationError` propagation on invalid payloads, and passthrough when no annotation is present.
- `ext/dapr-ext-workflow/tests/durabletask/test_serialization.py` — encoder tests: plain-dict emission (no `AUTO_SERIALIZED`), nested list of models, and Pydantic + dataclass coexistence (dataclass still marked, Pydantic not).

**Example (3 files):**
- `examples/workflow/pydantic_models.py` — new. End-to-end demo: schedule with a real `OrderRequest`, activity receives a typed instance, workflow reconstructs the activity result into a typed `OrderResult`, client calls `OrderResult.model_validate_json(state.serialized_output)` for a typed output.
- `examples/workflow/requirements.txt` — added `pydantic>=2.0`.
- `examples/workflow/README.md` — new "Pydantic models as workflow/activity inputs" section with a `<!--STEP-->` block so mechanical-markdown validates it in CI.

**Dev dependencies:** `pydantic>=2.0.0` was already present in `dev-requirements.txt`; comment updated to note workflow tests now rely on it.

**No changes** to: `setup.cfg` (no runtime dep, no extras), async client (parity is automatic because the encoder and receive-side introspection live upstream of the sync/async split), public API exports, protobuf.

## Test plan

- [x] `python -m pytest ext/dapr-ext-workflow/tests/` — 79 passing, no regressions
- [x] `ruff check` + `ruff format` clean
- [x] Example imports, registers workflow and activity, no runtime errors
- [ ] Run example end-to-end with a Dapr sidecar (`dapr run --app-id wf-pydantic-example -- python3 pydantic_models.py`) — please verify in CI / reviewer environment
- [ ] Verify mechanical-markdown picks up the new STEP block under `./validate.sh workflow`

## Edge cases covered

- `Optional[Model]` / `Model | None` annotations unwrap correctly.
- User passes an already-constructed model instance → passed through, no redundant re-validation.
- Invalid payload → Pydantic `ValidationError` surfaces cleanly through the activity error path with instance-id logging.
- No annotation → unchanged behavior (`SimpleNamespace` / dict).
- Primitive annotations (`int`, `str`, etc.) → passthrough, no coercion.
- Dataclasses and namedtuples continue to use `AUTO_SERIALIZED` round-tripping via `SimpleNamespace`; only the new protocol branch is marker-free.
- Duck-typed classes without a `mode` kwarg on `model_dump` use the bare-call fallback.
- Cross-app activities (`app_id='other-app'`) and cross-language callers are unaffected — the wire format is plain JSON with no SDK-specific markers.

### DIFF
--- a/ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
+++ b/ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
@@ -91,7 +91,7 @@ class WorkflowRuntime:
         effective_name = name or fn.__name__
         self._logger.info(f"Registering workflow '{effective_name}' with runtime")
 
-        input_model = _model_protocol.resolve_input_model(fn)
+        accepts_input, input_model = _model_protocol.resolve_input(fn)
 
         def orchestrationWrapper(ctx: task.OrchestrationContext, inp: Optional[TInput] = None):
             """Responsible to call Workflow function in orchestrationWrapper"""
@@ -99,10 +99,14 @@ class WorkflowRuntime:
 
             try:
                 daprWfContext = DaprWorkflowContext(ctx, self._logger.get_options())
-                if inp is None:
+                if not accepts_input:
                     result = fn(daprWfContext)
                 else:
-                    if input_model is not None and not isinstance(inp, input_model):
+                    if (
+                        inp is not None
+                        and input_model is not None
+                        and not isinstance(inp, input_model)
+                    ):
                         inp = _model_protocol.coerce_to_model(inp, input_model)
                     result = fn(daprWfContext, inp)
                 return result
@@ -137,14 +141,14 @@ class WorkflowRuntime:
             f"Registering version {version_name} of workflow '{effective_name}' with runtime"
         )
 
-        input_model = _model_protocol.resolve_input_model(fn)
+        accepts_input, input_model = _model_protocol.resolve_input(fn)
 
         def orchestrationWrapper(ctx: task.OrchestrationContext, inp: Optional[TInput] = None):
             """Responsible to call Workflow function in orchestrationWrapper"""
             daprWfContext = DaprWorkflowContext(ctx, self._logger.get_options())
-            if inp is None:
+            if not accepts_input:
                 return fn(daprWfContext)
-            if input_model is not None and not isinstance(inp, input_model):
+            if inp is not None and input_model is not None and not isinstance(inp, input_model):
                 inp = _model_protocol.coerce_to_model(inp, input_model)
             return fn(daprWfContext, inp)
 
@@ -177,7 +181,7 @@ class WorkflowRuntime:
         effective_name = name or fn.__name__
         self._logger.info(f"Registering activity '{effective_name}' with runtime")
 
-        input_model = _model_protocol.resolve_input_model(fn)
+        accepts_input, input_model = _model_protocol.resolve_input(fn)
 
         def activityWrapper(ctx: task.ActivityContext, inp: Optional[TInput] = None):
             """Responsible to call Activity function in activityWrapper"""
@@ -185,10 +189,14 @@ class WorkflowRuntime:
 
             try:
                 wfActivityContext = WorkflowActivityContext(ctx)
-                if inp is None:
+                if not accepts_input:
                     result = fn(wfActivityContext)
                 else:
-                    if input_model is not None and not isinstance(inp, input_model):
+                    if (
+                        inp is not None
+                        and input_model is not None
+                        and not isinstance(inp, input_model)
+                    ):
                         inp = _model_protocol.coerce_to_model(inp, input_model)
                     result = fn(wfActivityContext, inp)
                 return result
--- a/ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
+++ b/ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
@@ -19,6 +19,7 @@ from functools import wraps
 from typing import Optional, Sequence, TypeVar, Union
 
 import grpc
+from dapr.ext.workflow import _model_protocol
 from dapr.ext.workflow._durabletask import task, worker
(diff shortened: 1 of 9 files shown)


ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py
```diff
@@ -131,11 +141,15 @@ def register_versioned_workflow(
...
 
+        accepts_input, input_model = _model_protocol.resolve_input(fn)
+
         def orchestrationWrapper(ctx: task.OrchestrationContext, inp: Optional[TInput] = None):
             """Responsible to call Workflow function in orchestrationWrapper"""
             daprWfContext = DaprWorkflowContext(ctx, self._logger.get_options())
-            if inp is None:
+            if not accepts_input:
                 return fn(daprWfContext)
+            if inp is not None and input_model is not None and not isinstance(inp, input_model):
```
seherv, line 151: nitpick: I find it a bit hard to read chained conditions like these. Even if the priority is evident, I'd write this (and any other long conditions) as 
```python
if (inp is not None) and (input_model is not None) and not isinstance(inp, input_model):
```

CasperGN (approved):

PR #1020, opened 2026-05-07
Remove development builds and centralize versions into single file

### LINKED ISSUE #961: refactor: remove `-dev` packages
# Problem

Every package in this repo has a `setup.py` that renames the published package at build time:

- `dapr` → `dapr-dev`
- `dapr-ext-grpc` → `dapr-ext-grpc-dev`
- etc.

This runs on every push to main, publishing dev snapshots under separate PyPI package names. Extensions also swap their `dapr` dependency to `dapr-dev` in the same step.

This creates ongoing maintenance burden:
- Any new package needs this same boilerplate `setup.py`
- It's the only reason `setup.py` still exists — blocking a full migration to `uv build`/`uv publish`
- The logic is easy to get wrong (the workflow/strands extensions had it missing at one point)

## Why it's no longer needed

Python packaging already solves this natively. PEP 440 dev releases (`1.17.0.devN`) are excluded from the default resolver. `pip install dapr` will never install a `.dev` version — users would need to explicitly pass `--pre` or pin the exact version. This has been the case since pip 9 (2016).

Our version strings already follow the PEP 440 `.devN` format. We're just not taking advantage of it.

## Questions

- Are there any consumers of the `dapr-dev` / `*-dev` packages on PyPI?
- Any concern about dev releases appearing alongside stable releases on the `dapr` PyPI page? Potentially very noisy
- the `pip install --pre` will now install `.dev` packages and not just release candidates. Acceptable?

seherv:
# Description

Removes the `setup.py` files that created separate `-dev` builds for all packages, and updates docs to instruct users to install directly from source if they need the latest changes in main.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/python-sdk/issues/961

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


sicoyle:
pls resolve the conflicts @seherv

.github/workflows/build-tag.yaml
```diff
@@ -19,6 +19,27 @@ jobs:
...
+        "
+    - name: Verify release version is not a .dev
+      if: startsWith(github.ref, 'refs/tags/v')
+      run: |
+        uv run python -c "
+        from dapr.version import __version__
+        assert '.dev' not in __version__, (
+            f'__version__ {__version__!r} contains ".dev". Release tags must point to a stable or rc version'
+        )
+        "
```
CasperGN: Couldn't these be the same step?

CasperGN (changes requested):

ext/dapr-ext-fastapi/pyproject.toml
```diff
@@ -29,19 +29,16 @@ Documentation = "https://github.com/dapr/docs"
...
+path = "../../VERSION"
+pattern = '(?P<version>\S+)'
 
-[tool.setuptools.package-data]
-"dapr.ext.fastapi" = ["py.typed"]
-
-[tool.setuptools.dynamic]
-version = {attr = "dapr.ext.fastapi.version.__version__"}
+[tool.hatch.build.targets.wheel]
+only-include = ["dapr/ext/fastapi"]
```
sicoyle, line 41: can you share on the benefits of using hatchling instead here pls?

ext/flask_dapr/pyproject.toml
```diff
@@ -28,17 +28,16 @@ Documentation = "https://github.com/dapr/docs"
...
+path = "../../VERSION"
+pattern = '(?P<version>\S+)'
 
-[tool.setuptools.package-data]
-"flask_dapr" = ["py.typed"]
-
-[tool.setuptools.dynamic]
-version = {attr = "flask_dapr.version.__version__"}
+[tool.hatch.build.targets.wheel]
+only-include = ["flask_dapr"]
```
sicoyle, line 40: why is this one missing the dapr/ext prefix that the others include?

ext/flask_dapr/README.rst
```diff
@@ -20,17 +20,17 @@ Installation
...
+
+    from flask import Flask, request
+    from flask_dapr import DaprApp
+
+    app = Flask('myapp')
+    dapr_app = DaprApp(app)
+    @dapr_app.subscribe(pubsub='pubsub', topic='some_topic', route='/some_endpoint')
+    def my_event_handler():
+      # request.data contains pubsub event
+      pass
```
sicoyle, line 33: why? is this format the one that should be used for rst files?

pyproject.toml
```diff
@@ -120,16 +117,10 @@ warn_unused_configs = true
...
-    "dapr/conf/**/*.py",
-    "dapr/serializers/**/*.py",
-    "ext/dapr-ext-grpc/dapr/**/*.py",
-    "ext/dapr-ext-fastapi/dapr/**/*.py",
-    "ext/flask_dapr/flask_dapr/*.py",
-    "examples/demo_actor/**/*.py",
+namespace_packages = true
+packages = [
+    "dapr",
+    "flask_dapr",
```
sicoyle, line 123: is this missing any?

sicoyle (commented):
few comments - overall LGTM - thank you!!

ext/flask_dapr/README.rst
seherv: I did a test reupload of all the packages to my TestPyPI repo and it rejected them because the READMEs were not parsable by the rst library in strict mode 😓 

Doing this fixed it, and I don't know how it worked before, but rewriting everything in Markdown to avoid this in the future is an option too

ext/dapr-ext-fastapi/pyproject.toml
seherv: setuptools can't load the version number dynamically from a text file, and honestly I want to get rid of setuptools ASAP because it clashes a lot with modern `pyproject.toml` conventions.

Hatchling does have support for dynamic versioning and it's relatively well-known

ext/flask_dapr/pyproject.toml
seherv: The package structure is a bit different, all the other exts are organized as `dapr/ext/package/` but flask_dapr was the only one organized as just that, `flask_dapr/` as the top level package. The package in PyPI was also called that.

Ideally we'll remove these PyPI packages entirely when we bundle the extensions as PEP 771 extras to the SDK 😄

pyproject.toml
seherv: Hatchling is smarter than setuptools and is able to see all `dapr/ext/package`s as part of the dapr namespace, so they get auto included in the build. This one is the exception and we have to include it explicitly

acroca (commented):
LGTM! So much cleaner, thanks!
I'll let @sicoyle approve as the comments are still not resolved.

sicoyle (approved):
overall LGTM - thank you!!

acroca (approved):

CasperGN (approved):

Issue #85, opened 2020-07-09
Add scenario based application examples


youngbupark:
## Describe the proposal

Add more realistic application examples which can show the capabilities of the convenient layer.

wmeints:
Hi @youngbupark we've built a nice workshop at Info Support that demonstrates how to use Dapr in a more realistic scenario: [edwinvw/dapr-workshop](https://github.com/edwinvw/dapr-workshop). I've created [the python version]( https://github.com/wmeints/dapr-workshop-python) for that workshop.  I think it could be a great source for people to learn how to use Dapr in Python.

PR #1025, opened 2026-05-12
Add history propagation

acroca:
## Description

Adds workflow history propagation to `dapr-ext-workflow`.

A workflow can opt into propagating its execution history to a child workflow or activity via `propagation=PropagationScope.OWN_HISTORY` or `propagation=PropagationScope.LINEAGE` on `call_activity` / `call_child_workflow`.

The receiver reads the propagated chunk through `ctx.get_propagated_history()` and queries it with `PropagatedHistory.get_workflow_by_name(...)` → `WorkflowResult.get_activity_by_name(...)` / `.get_child_workflow_by_name(...)`.

Public surface lives in `dapr.ext.workflow.propagation` and is re-exported from `dapr.ext.workflow` — users never need to import from `_durabletask`. Default behavior is unchanged: `propagation=None` propagates nothing.

> Requires a Dapr sidecar built with history propagation support (durabletask-go #85+ / runtime 1.18+). Against an older sidecar the propagation field is silently dropped and `get_propagated_history()` returns `None`.

Includes an example app at `examples/workflow/history_propagation.py`.

## References

- Go SDK reference: https://github.com/dapr/durabletask-go/pull/85
- Proto changes: https://github.com/dapr/durabletask-protobuf/pull/38


examples/workflow/README.md
```diff
@@ -536,3 +536,38 @@ timeout_seconds: 60
...
+<!--STEP
+name: Run the history propagation example
+expected_stdout_lines:
+  - "*** validating merchant merchant-42"
+  - "*** process_payment received parent context for merchant 'merchant-42'"
+  - "*** log_summary saw parent on app"
+  - "validate_merchant -> completed=True output={\"merchant_id\": \"merchant-42\", \"valid\": true}"
+  - "*** workflow completed: status=COMPLETED"
+timeout_seconds: 60
+-->
```
seherv: I think we can skip the Mechanical Markdown now

ext/dapr-ext-workflow/dapr/ext/workflow/propagation.py
```diff
@@ -0,0 +1,349 @@
...
+        Each chunk's ``rawEvents`` are parsed once and the per-chunk events are
+        concatenated into a single ordered list. Structural validation runs
+        first: every chunk must carry a non-empty ``appId`` and every raw event
+        must parse as a ``HistoryEvent``. Returns ``None`` when ``ph`` itself
+        is ``None``.
+
+        Raises:
+            ValueError: If the proto is structurally malformed.
+        """
+        if ph is None:
```
seherv: Can you please rename this variable to something like `propagated_history`? It's easy to confuse with the protobuf stuff (`pb` and `pbh`)

ext/dapr-ext-workflow/dapr/ext/workflow/propagation.py
```diff
@@ -0,0 +1,349 @@
...
+        """All workflow results in the chain, in execution order
+        (ancestor first, immediate parent last)."""
+        return [self._make_workflow_result(c) for c in self._chunks]
+
+    def get_workflows_by_name(self, name: str) -> list[WorkflowResult]:
+        """All workflows whose name matches, in execution order. Useful when
+        the chain contains the same name more than once (recursion / ContinueAsNew)."""
+        return [self._make_workflow_result(c) for c in self._chunks if c.workflow_name == name]
+
+    def get_workflow_by_name(self, name: str) -> WorkflowResult:
```
seherv, line 309: I know we're keeping parity with durabletask-go but I had to read these function names twice 👀 

Assuming the API is completely frozen by now, maybe it would help if the docstring of one of these two functions mentioned the other function?

ext/dapr-ext-workflow/dapr/ext/workflow/propagation.py
```diff
@@ -0,0 +1,349 @@
...
+            start = len(events)
+            for j, raw in enumerate(c.rawEvents):
+                event = pb.HistoryEvent()
+                try:
+                    event.ParseFromString(raw)
+                except Exception as ex:
+                    raise ValueError(
+                        f'propagated history: chunk {i} (app {c.appId!r}): '
+                        f'failed to decode rawEvent {j}: {ex}'
+                    ) from ex
```
seherv, line 369: The way I am reading this, a single unparsable event would crash the entire workflow. If that is the case, is that the expected behavior?

ext/dapr-ext-workflow/dapr/ext/workflow/propagation.py
```diff
@@ -0,0 +1,349 @@
...
+            ValueError: If the proto is structurally malformed.
+        """
+        if ph is None:
+            return None
+
+        events: list[pb.HistoryEvent] = []
+        chunks: list[_HistoryChunk] = []
+        for i, c in enumerate(ph.chunks):
+            if not c.appId:
+                raise ValueError(f'propagated history: chunk {i} has empty appId')
```
seherv, line 359: Should the instance ID and the workflow names be checked too, or are those not critical?

ext/dapr-ext-workflow/dapr/ext/workflow/propagation.py
```diff
@@ -0,0 +1,349 @@
...
+
+    @property
+    def events(self) -> list[pb.HistoryEvent]:
+        """All propagated history events, flattened in chunk order."""
+        return self._events
+
+    @property
+    def scope(self) -> PropagationScope:
+        """The propagation scope used to produce this history."""
+        return self._scope
```
seherv: The `@property` is making these variables effectively public. 

I'd remove the `@property` and the underscore for `scope`. 
For `events` I'd either make it return an immutable `tuple(self.events)` or make it fully public just like `scope` (if mutability is not a problem)

seherv (changes requested):

ext/dapr-ext-workflow/dapr/ext/workflow/propagation.py
acroca: Yes, intentional. The user should trust what they receive, so if it's malformed we have to error out. They can always capture this individual exception and act accordingly.
I'll extend the docs to be clearer
What do you think?

ext/dapr-ext-workflow/dapr/ext/workflow/propagation.py
acroca: Only appID is required. Extended the docs.

seherv (approved):

CasperGN (approved):
LGTM @acroca

Issue #1001, opened 2026-04-30
Add support for workflow history propagation


cicoyle:
## Background

Dapr Workflows now supports propagating execution history from a parent workflow to its child workflows and activities. This lets downstream code inspect what happened upstream — useful for chain-of-custody verification, fraud detection, audit, and AI-agent context flow.

Two scopes are exposed:

- **`PropagateLineage`** — caller's own events plus the full ancestor chain
- **`PropagateOwnHistory`** — caller's events only (a trust boundary; ancestors are dropped)

The Go path is implemented end-to-end. To use this feature from a Python app, this SDK needs the matching API.

## Proto surface

From `dapr/durabletask-protobuf` ([PR #38](https://github.com/dapr/durabletask-protobuf/pull/38)):

- `history_events.proto`
  - `TaskScheduledEvent.historyPropagationScope` (optional, field 7)
  - `ChildWorkflowInstanceCreatedEvent.historyPropagationScope` (optional, field 7)
  - new `PropagatedHistory { repeated HistoryEvent events; HistoryPropagationScope scope; repeated PropagatedHistoryChunk chunks; }`
  - new `PropagatedHistoryChunk { string appId; int32 startEventIndex; int32 eventCount; string instanceId; string workflowName; }`
- `orchestration.proto`
  - new `HistoryPropagationScope` enum: `HISTORY_PROPAGATION_SCOPE_NONE = 0`, `HISTORY_PROPAGATION_SCOPE_OWN_HISTORY = 1`, `HISTORY_PROPAGATION_SCOPE_LINEAGE = 2`
- `orchestrator_actions.proto`
  - `ScheduleTaskAction.historyPropagationScope` (optional, field 6)
  - `CreateChildWorkflowAction.historyPropagationScope` (optional, field 6)

### 1. Proto regen + plumbing

- Pull the updated `.proto` files from `dapr/durabletask-protobuf` and regenerate the Python protobuf classes
- Plumb the new `historyPropagationScope` field through the schedule-task / create-child-workflow code paths
- Surface the received `PropagatedHistory` to the workflow runtime so it can be handed to user code

### 2. User-facing API to add

**Schedule-side (parent workflow author):**
- `with_history_propagation(scope)` keyword/option on `call_activity` / `call_child_workflow`
[...]

PR #1047, opened 2026-05-20
feat(workflow): align history propagation API with go-sdk

nelson-parente:
## Summary

Aligns the workflow history propagation surface added in #1025 with the freshly-renamed go-sdk helpers ([durabletask-go#105](https://github.com/dapr/durabletask-go/pull/105), merged 2026-05-19). That rename landed four days after #1025 merged, so this is drift, not a defect — Cassie flagged it post-merge while building the cross-SDK quickstarts.

This brings python-sdk back in line for cross-SDK parity before 1.18 ships.

### Read API renames

Mirror durabletask-go's `GetLast*ByName` form to make it obvious these return the most-recent occurrence (with plural `get_*s_by_name` siblings for the full list):

| Old | New |
| --- | --- |
| `PropagatedHistory.get_workflow_by_name` | `get_last_workflow_by_name` |
| `WorkflowResult.get_activity_by_name` | `get_last_activity_by_name` |
| `WorkflowResult.get_child_workflow_by_name` | `get_last_child_workflow_by_name` |

Plural variants (`get_workflows_by_name`, `get_activities_by_name`, `get_child_workflows_by_name`) and chain-level helpers (`get_app_ids`, `get_events_by_*`) are unchanged.

### Scheduling — use the enum directly

The propagation scope is passed via the existing `propagation=` kwarg, using `PropagationScope` members:

```python
import dapr.ext.workflow as wf

yield ctx.call_child_workflow(
    process_payment,
    input=req,
    propagation=wf.PropagationScope.OWN_HISTORY,
)

yield ctx.call_activity(
    log_summary,
    input=None,
    propagation=wf.PropagationScope.LINEAGE,
)
```

> An earlier revision of this PR shipped `propagate_lineage()` / `propagate_own_history()` factory functions mirroring the go-sdk shape. Per review feedback from @seherv these were dropped (commit 69a78d6) — Go uses factories to compensate for the lack of enums; Python has them, so the enum members are clearer and play better with type checkers.

### Non-goals

- **Kwarg name unchanged** (`propagation=`). Renaming to `with_history_propagation=` would mirror go-sdk's `WithHistoryPropagation()` more literally but isn't idiomatic Python.
- **No proto / runtime changes.** This PR is purely surface-level: renamed query methods are direct renames.

## Test plan

- [x] `uv run ruff check --fix && uv run ruff format` — clean
- [x] `uv run pytest ext/dapr-ext-workflow/tests/durabletask/test_propagation.py ext/dapr-ext-workflow/tests/durabletask/test_propagation_wiring.py` — 29 passed
- [x] `uv run python -m unittest discover -v ./ext/dapr-ext-workflow/tests` — 133 passed
- [ ] `uv run pytest tests/examples/test_workflow.py::test_history_propagation` against a 1.18-RC sidecar (output-based test; print statements unchanged, should still pass)
- [ ] CI green

## References

- Closes (partially) #1001
- go-sdk reference rename: dapr/durabletask-go#105
- go-sdk reference impl + example: dapr/go-sdk#823
- java-sdk parallel: dapr/java-sdk#1739

cc @cicoyle @acroca

### DIFF
--- a/examples/workflow/README.md
+++ b/examples/workflow/README.md
@@ -544,12 +544,12 @@ history to a child workflow and to an activity, and how the receivers query
 that history through `ctx.get_propagated_history()`.
 
 It shows:
-- `propagation=PropagationScope.OWN_HISTORY` on a child workflow call —
+- `propagation=propagate_own_history()` on a child workflow call —
   forwards the caller's events only.
-- `propagation=PropagationScope.LINEAGE` on an activity call — forwards the
+- `propagation=propagate_lineage()` on an activity call — forwards the
   caller's events *plus* anything the caller itself received from its parent.
-- `PropagatedHistory.get_workflow_by_name(...)` and `WorkflowResult.get_activity_by_name(...)`
-  on the receiving side.
+- `PropagatedHistory.get_last_workflow_by_name(...)` and
+  `WorkflowResult.get_last_activity_by_name(...)` on the receiving side.
 
 > **Requires** a Dapr sidecar with workflow history propagation support
 > (durabletask-go PR #85 / runtime 1.18+ ). With an older sidecar the
--- a/examples/workflow/README.md
+++ b/examples/workflow/README.md
@@ -544,9 +544,9 @@ history to a child workflow and to an activity, and how the receivers query
 that history through `ctx.get_propagated_history()`.
 
 It shows:
-- `propagation=propagate_own_history()` on a child workflow call —
+- `propagation=PropagationScope.OWN_HISTORY` on a child workflow call —
   forwards the caller's events only.
-- `propagation=propagate_lineage()` on an activity call — forwards the
+- `propagation=PropagationScope.LINEAGE` on an activity call — forwards the
   caller's events *plus* anything the caller itself received from its parent.
 - `PropagatedHistory.get_last_workflow_by_name(...)` and
   `WorkflowResult.get_last_activity_by_name(...)` on the receiving side.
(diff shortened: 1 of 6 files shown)


examples/workflow/README.md
```diff
@@ -544,12 +544,12 @@ history to a child workflow and to an activity, and how the receivers query
 that history through `ctx.get_propagated_history()`.
 
 It shows:
-- `propagation=PropagationScope.OWN_HISTORY` on a child workflow call —
+- `propagation=propagate_own_history()` on a child workflow call —
   forwards the caller's events only.
-- `propagation=PropagationScope.LINEAGE` on an activity call — forwards the
+- `propagation=propagate_lineage()` on an activity call — forwards the
   caller's events *plus* anything the caller itself received from its parent.
```
seherv: The way this is implemented would work fine, but IMO this is not idiomatic Python. 

It also obscures things a bit (and might even mess with type checkers) because the return type of these functions only tells us that `propagation=OwnHistory|Lineage`, while the plain enum value tells us exactly what the value is without a doubt.

If I understand correctly, this is the way Go makes up for a lack of enums and is not a pattern we should apply in Python.

seherv (changes requested):
Thanks for opening the PR! Left you one comment

PD: I agree with keeping it as `propagation=`

examples/workflow/README.md
nelson-parente: ty! Let me work around this feedback!

cicoyle (approved):

sicoyle (approved):

marcduiker:
@holopin-bot @nelson-parente sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

Issue #975, opened 2026-04-02
Investigate Python SDK concurrency limitations


sicoyle:
The Python Dapr SDK currently relies on threadpool-based execution for workflow and activity processing. This model has known limitations:

- The GIL (Global Interpreter Lock) constrains true parallelism for CPU-bound work
- Thread pool exhaustion under high concurrency can delay sidecar acknowledgments

Native asyncio execution would allow the SDK to handle a higher volume of concurrent workflow tasks without the overhead and contention introduced by thread management, reducing sidecar response latency under sustained load.

Goals

- Quantify the throughput and latency characteristics of the current threadpool implementation under sustained workflow load
- Rewrite workflow and activity execution to use native asyncio
- Validate that the rewrite improves sidecar response latency and throughput under load conditions

Tasks

 - [ ] Benchmark current threadpool implementation — measure workflow throughput, activity concurrency, and sidecar acknowledgment latency under sustained load
 - [ ] Document failure conditions — identify load thresholds at which sidecar response times begin to degrade
 - [ ] Design asyncio execution model — propose the replacement architecture for workflow and activity dispatch
 - [ ] Implement asyncio rewrite — replace threadpool-based execution with native asyncio in the workflow worker
 - [ ] Ensure existing workflow and activity behavior is preserved
 - [ ] Load validate — re-run benchmarks against the rewritten implementation under production-representative load and confirm improvement in sidecar response latency and throughput
 - [ ] Document operational guidance — update SDK documentation with concurrency configuration recommendations for workflow-heavy deployments
 - [ ] Write Dapr OSS blog on findings

olitomlinson:
related https://github.com/dapr/python-sdk/issues/897

olitomlinson:
bump

Issue #897, opened 2026-01-27
[WORKFLOW SDK BUG] Activity concurrency is throttled too much


olitomlinson:
**runtime** 
1.17.rc2

**SDK** 
dapr>=1.17.0rc2
dapr-ext-workflow>=1.17.0rc2

## Expected Behavior

Activities are allowed to run concurrently

## Actual Behavior

The SDK is throttling how many activities can be ran at any one point time

## Steps to Reproduce the Problem

Create a workflow, fan out 100 instances of Activities, have each activity sleep for 60 seconds before completing.

Look at the logs and notice that there are at most 10-15 activities running at the same time. Despite all Activities being **scheduled** at the same time, Activities won't **Activate** untill previous activities have **finished**.

Look at the timestamps in the logs below

note : When comparing this exact same workflow authored in dotnet, dotnet allows all activities to run concurrently.

```
2026-01-27 12:49:47


2026-01-27 12:49:47,810 - hello_world_app.workflows.fan_out_workflow - INFO - Starting fan-out workflow with 100 iterations
2026-01-27 12:49:47


2026-01-27 12:49:47.812 durabletask-worker INFO: 0-312: Orchestrator yielded with 100 task(s) and 0 event(s) outstanding.
2026-01-27 12:49:47


2026-01-27 12:49:47,812 - hello_world_app.workflows.fan_out_workflow - INFO - Waiting for 100 activities to complete
2026-01-27 12:49:47


2026-01-27 12:49:47,812 - durabletask-worker - INFO - 0-312: Orchestrator yielded with 100 task(s) and 0 event(s) outstanding.
2026-01-27 12:49:47


2026-01-27 12:49:47,824 - hello_world_app.activities.super_slow_activity - INFO - 0-312 - super_slow_activity (1) #0 - scheduled=12:49:47 activated=12:49:47
2026-01-27 12:49:47


2026-01-27 12:49:47,826 - hello_world_app.activities.super_slow_activity - INFO - 0-312 - super_slow_activity (1) #1 - scheduled=12:49:47 activated=12:49:47
2026-01-27 12:49:47


2026-01-27 12:49:47,826 - hello_world_app.activities.super_slow_activity - INFO - 0-312 - super_slow_activity (1) #2 - scheduled=12:49:47 activated=12:49:47
2026-01-27 12:49:47


[...]

PR #1053, opened 2026-05-25
Add support for async workflow activities

### LINKED ISSUE #834: feat: asyncio support for workflow sdk

seherv:
# Description

Workflow activities can now be async, and the runtime will automatically dispatch them to the event loop. Sync activities are still dispatched to the thread pool. The user-facing API remains exactly the same.

## Basic benchmarks

Ran all of these on an M3 Pro (Python 3.13). The async activities take a 100KB array of random data, do `asyncio.sleep(1)` and return the zeroes unmodified.

### 1. Fan-out from one workflow
  | Activities | Async |
  | ---: | ---: | 
  | 100 | 1.03 s | 
  | 1 000 | 1.24 s | 
  | 10 000 | 11.03 s | 

  ### 2. Fan-out from multiple workflows × 3 activities
  | Workflows × activities | Async | 
  | ---: | ---: | 
  | 100 × 3 | 1.09 s |
  | 1 000 × 3 | 3.60 s | 
  | 10 000 × 3 | 39.51 s | 

  ### 3. Variable payload (100 workflows × 3 activities)  
  | Payload | Async |
  | ---: | ---: |
  | 1 KB | 3.23 s |
  | 100 KB | 3.64 s |
  | 1 MB | 8.80 s |

## Issue reference

Closes https://github.com/dapr/python-sdk/issues/834, https://github.com/dapr/python-sdk/issues/897 and https://github.com/dapr/python-sdk/issues/975

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


dapr/ext/workflow/_durabletask/worker.py
```diff
@@ -964,98 +981,176 @@ def _execute_orchestrator(
...
-                stub.CompleteActivityTask(res)
-            except grpc.RpcError as rpc_error:  # type: ignore
-                if _is_message_too_large(rpc_error):
-                    # Result is too large to deliver - fail the activity immediately.
-                    # This can only be fixed with infrastructure changes (increasing gRPC max message size).
-                    self._logger.error(
-                        f"Activity '{req.name}#{req.taskId}' result is too large to deliver "
-                        f'(RESOURCE_EXHAUSTED). Failing the activity task: {rpc_error.details()}'
-                    )
-                    failure_res = pb.ActivityResponse(
```
seherv, line 1031: This nesting got hard to follow, I needed to refactor this file to understand the logic better.

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```diff
@@ -0,0 +1,1457 @@
...
+        submit_ts[i] = time.perf_counter()
+        manager.submit_activity(handler, req, stub, '')
+
+    await manager.activity_queue.join()
+    wallclock_s = time.perf_counter() - submit_start
+
+    manager._shutdown = True
+    sampler.stop()
+    await asyncio.gather(worker_task, sampler_task, return_exceptions=True)
+    manager.shutdown()
```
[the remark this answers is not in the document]
seherv: Not a problem, benchmarks are internal anyway.

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```diff
@@ -0,0 +1,1457 @@
...
+    while manager.activity_queue is None:
+        await asyncio.sleep(0)
+    sampler.watch_queue(manager.activity_queue)
+
+    submit_ts: dict[int, float] = {}
+    submit_start = time.perf_counter()
+    for i in range(n_items):
+        req = _build_activity_request(activity_name, task_id=i, instance_id='bench')
+        submit_ts[i] = time.perf_counter()
+        manager.submit_activity(handler, req, stub, '')
```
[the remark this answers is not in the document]
seherv: Forgot to push all changes, fixed

dapr/ext/workflow/workflow_runtime.py
```diff
@@ -45,6 +46,58 @@
     grpc.StreamStreamClientInterceptor,
 ]
 
+# Durabletask returns decoded JSON, so we type the input as ``object | None`` and let the
+# wrapper narrow it via the activity's declared model.
+SyncActivityWrapper = Callable[[task.ActivityContext, object | None], object]
+AsyncActivityWrapper = Callable[[task.ActivityContext, object | None], Awaitable[object]]
+ActivityWrapper = SyncActivityWrapper | AsyncActivityWrapper
```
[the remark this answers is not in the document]
seherv: Project is clearly 3.10+ from pyproject.toml

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/shared.py
```diff
@@ -19,6 +21,28 @@
...
+    candidate = fn
+    while isinstance(candidate, functools.partial):
+        candidate = candidate.func
+    if callable(candidate):
+        try:
+            candidate = inspect.unwrap(candidate)
+        except ValueError:
+            # Cyclic ``__wrapped__`` chain from a malformed decorator. Fall back to the
+            # outermost callable; misclassification is preferable to crashing dispatch.
+            pass
```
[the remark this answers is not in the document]
seherv: Tested and proven to be false in 3.10-3.13.

dapr/ext/workflow/_durabletask/worker.py
```diff
@@ -964,98 +978,179 @@ def _execute_orchestrator(
...
+                    res,
+                    completionToken,
+                    instance_id,
+                )
+            except RuntimeError as exc:
+                # Only swallow if the manager thread pool was shut down (worker is tearing
+                # down). The sidecar will re-dispatch the work item once the worker reconnects.
+                # Other RuntimeErrors are unexpected and bubble up to the work-item processor.
+                if not self._async_worker_manager._shutdown:
+                    raise
```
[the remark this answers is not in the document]
seherv: Won't do, it's an internal module accessing internal stuff

seherv:
Improving benchmarks and regression tests before submitting for review again

acroca:
Thanks for working on this! I agree with the general direction, async activity support is something we've wanted for a while (#834, #897, #975), and keeping the user-facing API unchanged is the right approach.

I'd like to share my opinion on the benchmarking side, though. I think the sync-vs-async comparison is valuable in the context of *this PR*, to justify the why behind the change, but not as something we keep and maintain in the codebase. Concretely:

- The benchmark suite adds ~1,100 lines (`benchmarks/`, `_bench_harness.py`), which is more code than the feature itself, and the comparison is essentially a one-off: once async dispatch is merged, it has served its purpose.
- `_bench_harness.py` sits inside `dapr/ext/workflow/`, so it would ship to every user as part of the package.
- The regression tests in `test_async_dispatch_regression.py` mostly assert sync-vs-async ratios, which again is the one-off comparison, and they depend on the harness.

My suggestion: run the benchmarks locally and paste the results + methodology into the PR description. Then the harness, the `benchmarks/` directory, and the ratio-based tests can be dropped from the PR.

That said, I do think we should keep one simple throughput regression test so the async path can't silently regress, something like running 1000 activities that each take 1 second and asserting the batch completes in well under 10 seconds (ideally it's ~1s, so a 10x bound leaves plenty of headroom for CI noise). That single assertion catches the failure mode we actually care about, async activities accidentally getting serialized, without needing the full harness, and it can be written directly against the runtime in a few dozen lines.

Thoughts?

ext/dapr-ext-workflow/benchmarks/_report.py
```diff
@@ -0,0 +1,223 @@
...
+from pathlib import Path
+
+from dapr.ext.workflow._bench_harness import IS_DARWIN, ScenarioMetrics, SustainedMetrics
+
+
+def _read_text(path: str) -> str:
+    try:
+        return Path(path).read_text(encoding='utf-8', errors='ignore')
+    except OSError:
+        return ''
```
sicoyle: could you maybe add a warn log here

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```diff
@@ -0,0 +1,353 @@
...
+production dispatch path against a mock sidecar stub. Scenarios: a fan-out burst, a
+fan-out shaped as many small workflows, and a sustained open-loop run.
+
+Run:
+
+    uv run python ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
+
+``DAPR_BENCH_ACTIVITY_MS`` overrides the activity duration, ``DAPR_BENCH_SUSTAINED_SECONDS``
+the sustained run. Writes ``benchmarks/RESULTS.md`` and asserts pass-criteria budgets.
+"""
```
sicoyle: ideally this is not needed and is included in a readme pls

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```diff
@@ -0,0 +1,353 @@
...
+async def run_by_scale() -> list[Row]:
+    """Varying workflows x activities, fixed payload size."""
+    rows: list[Row] = []
+    for w in WF_ACTIVITY_GRID:
+        label = f'{w} × {ACTIVITIES_PER_WORKFLOW}'
+        rows.append(await _workflow_pair(label, w * ACTIVITIES_PER_WORKFLOW, FIXED_PAYLOAD_BYTES))
+    return rows
+
+
+async def run_sustained() -> tuple[SustainedMetrics, SustainedMetrics]:
```
sicoyle: what does this mean sustained? like constant throughput over the same duration?

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```diff
@@ -0,0 +1,353 @@
...
+    ceiling = _THREAD_POOL / ACTIVITY_LATENCY_S
+    latency_ms = ACTIVITY_LATENCY_S * 1000
+    fixed_kb = FIXED_PAYLOAD_BYTES // 1024
+    report = f"""
+# Sync vs async activity benchmark
+
+Generated by `bench_async_activities.py`. Re-run with:
+
+```bash
+uv run python ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```
sicoyle: pls don't include how to run here bc this can easily become stale if we rename the file or move things.

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```diff
@@ -0,0 +1,353 @@
...
+
+{_format_comparison_table(by_payload, key_label='Payload', show_async_rss=True)}
+
+## 3. Sustained arrival
+
+Open-loop submission at {rate:.0f}/s, above the ~{ceiling:.0f}/s sync ceiling. The sync path falls behind and its latency drifts upward across the run; the async path holds steady. Compare the first-quarter and last-quarter p99.
+
+{_format_sustained_table(sustained_sync, sustained_async)}
+
+See `ext/dapr-ext-workflow/docs/concurrency.md` for sizing guidance.
```
sicoyle: is this a doc you plan to share? or can you rm this ref pls?

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
```diff
@@ -0,0 +1,353 @@
...
+"""
+    RESULTS_PATH.write_text(report, encoding='utf-8')
+
+
+def _assert_budgets(
```
sicoyle: can you find a more indicative word instead of budgets pls? i see this checks a mix of wallclock and other data so maybe split to a few assert funcs or something to be cleaner?

dapr/ext/workflow/_durabletask/aio/internal/shared.py
```diff
@@ -42,6 +67,8 @@ def get_grpc_aio_channel(
         interceptors: Optional sequence of client interceptors to apply to the channel.
         options: Optional sequence of gRPC channel options as (key, value) tuples. Keys defined in https://grpc.github.io/grpc/core/group__grpc__arg__keys.html
     """
+    _silence_grpc_aio_poller_noise()
```
sicoyle, line 71: is this only on the asyncio side of things?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/internal/shared.py
sicoyle: should we warn log here?

dapr/ext/workflow/_durabletask/internal/shared.py
```diff
@@ -19,6 +21,28 @@
 import grpc
 from dapr.ext.workflow import _model_protocol
 
+
+def is_async_callable(fn: Any) -> bool:
```
sicoyle, line 28: is there a test somewhere for this addition?

dapr/ext/workflow/_durabletask/aio/internal/shared.py
```diff
@@ -27,6 +28,30 @@
...
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        exc = record.exc_info[1] if record.exc_info else None
+        is_poller_noise = isinstance(exc, BlockingIOError) and (
+            _POLLER_NOISE_MARKER in record.getMessage()
+        )
+        return not is_poller_noise
+
+
+def _silence_grpc_aio_poller_noise() -> None:
```
sicoyle, line 50: add a test pls

dapr/ext/workflow/_durabletask/worker.py
```diff
@@ -964,98 +979,178 @@ def _execute_orchestrator(
...
+            return contextlib.nullcontext()
+        return otel_tracer.start_as_current_span(
+            name=f'activity: {req.name}',
+            context=otel_propagator.extract(
+                carrier={'traceparent': req.parentTraceContext.traceParent}
+            ),
+            attributes={
+                'dapr.ext.workflow._durabletask.task.instance_id': instance_id,
+                'dapr.ext.workflow._durabletask.task.id': req.taskId,
+                'dapr.ext.workflow._durabletask.activity.name': req.name,
```
sicoyle, line 995: were these always underscore as the durabletask prefix?

ext/dapr-ext-workflow/dapr/ext/workflow/_durabletask/worker.py
```diff
@@ -30,6 +30,7 @@
 import grpc
```
sicoyle: pls add tests for all the helpers you created in this file 🙏

ext/dapr-ext-workflow/dapr/ext/workflow/_bench_harness.py
```diff
@@ -0,0 +1,541 @@
...
+        return rss // 1024
+    return rss
+
+
+# ============================================================================
+# Mock sidecar stub (production response path goes through here)
+# ============================================================================
+
+
+class _MockSidecarStub:
```
sicoyle: can you mv to a diff file for the mock pls

ext/dapr-ext-workflow/dapr/ext/workflow/_bench_harness.py
```diff
@@ -0,0 +1,541 @@
...
+def _activity_and_handler(
+    worker: TaskHubGrpcWorker,
+    kind: str,
+    factory: ActivityFactory | None,
+    latency_s: float,
+    start_ts: dict[int, float],
+    end_ts: dict[int, float],
+) -> tuple[Callable[..., object], Callable[..., object]]:
+    """Build the activity callable and pick the matching dispatch handler for ``kind``."""
+    if kind == 'async':
```
sicoyle: const pls

ext/dapr-ext-workflow/dapr/ext/workflow/_bench_harness.py
```diff
@@ -0,0 +1,541 @@
...
+        fn = (
+            factory(start_ts, end_ts)
+            if factory
+            else _async_sleep_factory(latency_s, start_ts, end_ts)
+        )
+        return fn, worker._execute_activity_async
+    if kind == 'sync':
```
sicoyle: const and switch pls

ext/dapr-ext-workflow/dapr/ext/workflow/_bench_harness.py
```diff
@@ -0,0 +1,541 @@
...
+    """Build a TaskHubGrpcWorker without calling start(). We only need its dispatch
+    code and registry; the gRPC stream is replaced by the mock stub.
+    """
+    return TaskHubGrpcWorker(
+        host_address='in-process-mock',
+        concurrency_options=options,
+    )
+
+
+ActivityFactory = Callable[[dict[int, float], dict[int, float]], Callable[..., object]]
```
sicoyle: in general can you also split this file up with smaller appropriately named files pls?

ext/dapr-ext-workflow/dapr/ext/workflow/_bench_harness.py
```diff
@@ -0,0 +1,541 @@
...
+        throughput_per_s=completed / wallclock_s if wallclock_s > 0 else 0.0,
+        latency=LatencyStats.from_samples(e2e_samples),
+        peak_tasks=sampler.peak_tasks,
+        peak_queue_depth=sampler.peak_queue_depth,
+        peak_rss_delta_mb=max(0.0, (sampler.peak_rss_kb - baseline_rss_kb) / 1024.0),
+    )
+
+
+def _make_activity_context(orchestration_id: str, task_id: int) -> task.ActivityContext:
+    return task.ActivityContext(orchestration_id, task_id, '', propagated_history=None)
```
sicoyle: is it really worth creating helpers that are one liners?

ext/dapr-ext-workflow/dapr/ext/workflow/_bench_harness.py
```diff
@@ -0,0 +1,541 @@
...
+        thread_pool_workers=thread_pool_workers,
+        server_latency_s=server_latency_s,
+        wallclock_s=wallclock_s,
+        e2e_samples=e2e,
+        sampler=sampler,
+        baseline_rss_kb=baseline_rss_kb,
+    )
+
+
+async def _run_lite(
```
sicoyle: what is lite meaning here?

sicoyle: maybe docstrings would help on some of these funcs pls?

sicoyle (commented):
few comments so far. I'm almost half way through the files/changes :)

ext/dapr-ext-workflow/benchmarks/bench_async_activities.py
seherv: Sustained as in constant flow of incoming activities every few ms during a fixed time period. It's there to show that the thread pool gets overloaded and that the async version can take it just fine

dapr/ext/workflow/_durabletask/internal/shared.py
seherv: Yup, it's at `tests/test_async_activity_registration.py`

seherv:
Your comments have convinced me that the repo is not the right place for benchmark code. It was very useful to validate everything myself, but I think our users only need to see a summary of the results and their applications are the real stress tests anyway.

I'll remove the bench code and simplify the regression tests to validate only the async side of things.

dapr/ext/workflow/_durabletask/aio/internal/shared.py
seherv: Yup, grpc.aio spams the error logs when their client is used from multiple event loops, and that was the case for FastAPI applications using this SDK.  Nothing was actually an error but the logs got extremely noisy in Linux.

It got fixed on their 1.80.0 release, as soon as we update to that dep (in a separate PR ofc) we can delete this

ext/dapr-ext-workflow/tests/durabletask/test_async_dispatch_regression.py
```diff
@@ -0,0 +1,104 @@
...
+
+import asyncio
+
+import dapr.ext.workflow._durabletask.internal.protos as pb
+import pytest
+from dapr.ext.workflow._durabletask.worker import ConcurrencyOptions, TaskHubGrpcWorker
+
+pytestmark = pytest.mark.perf
+
+ACTIVITY_S = 0.02
```
sicoyle: what is activity s?

tests/ext/workflow/durabletask/test_async_dispatch_regression.py
```diff
@@ -0,0 +1,104 @@
...
+    worker = TaskHubGrpcWorker(host_address='in-process-mock', concurrency_options=options)
+    manager = worker._async_worker_manager
+    stub = _MockSidecarStub()
+
+    async def activity(ctx, _inp) -> None:
+        await asyncio.sleep(ACTIVITY_S)
+
+    worker_task = asyncio.create_task(manager.run())
+    while manager.activity_queue is None:
+        await asyncio.sleep(0)
```
sicoyle, line 77: what does sleep 0 do?

uv.lock
```diff
@@ -1467,7 +1467,7 @@ wheels = [
 
 [[package]]
 name = "langsmith"
-version = "0.8.6"
+version = "0.7.17"
```
sicoyle: why the downgrade here?

uv.lock
```diff
@@ -2611,11 +2610,11 @@ wheels = [
 
 [[package]]
 name = "python-multipart"
-version = "0.0.29"
+version = "0.0.22"
```
sicoyle: yeah, we're downgrading a few things can you pls revert this change?

ext/dapr-ext-workflow/AGENTS.md
```diff
@@ -105,6 +105,26 @@ The entry point for registration and lifecycle:
...
+- **Sync activities** go through `_execute_activity`, dispatched to the thread pool by `_AsyncWorkerManager._run_func`. The activity runs on a worker thread, and the response is delivered from the same thread.
+
+Workflow (orchestrator) functions must remain generators (`def` with `yield`). They cannot be `async def` because durabletask's deterministic replay depends on synchronous generator semantics. Only activities support async.
+
+**Decorator ordering gotcha.** Stacking `@wfr.activity` over `@alternate_name(...)` over `async def` works because `@alternate_name` now emits an `async def innerfn` when the wrapped function is async. A user-written decorator that wraps an async function in a sync `def` (without `@functools.wraps` exposing `__wrapped__`) defeats `_is_async_callable`, routes the activity to the sync path, and produces an un-awaited coroutine. Such decorators should use `@functools.wraps(fn)` so the unwrap walks through them.
```
sicoyle: what does this part mean?

sicoyle: oh, maybe change Stacking -> Wrapping?

sicoyle: based on the unawaited coroutine part, do we capture this error and bubble up a nice msg in the event that this occurs?

dapr/ext/workflow/AGENTS.md
```diff
@@ -105,6 +105,26 @@ The entry point for registration and lifecycle:
...
+
+Workflow (orchestrator) functions must remain generators (`def` with `yield`). They cannot be `async def` because durabletask's deterministic replay depends on synchronous generator semantics. Only activities support async.
+
+**Decorator ordering gotcha.** Stacking `@wfr.activity` over `@alternate_name(...)` over `async def` works because `@alternate_name` now emits an `async def innerfn` when the wrapped function is async. A user-written decorator that wraps an async function in a sync `def` (without `@functools.wraps` exposing `__wrapped__`) defeats `_is_async_callable`, routes the activity to the sync path, and produces an un-awaited coroutine. Such decorators should use `@functools.wraps(fn)` so the unwrap walks through them.
+
+**`maximum_thread_pool_workers` covers both paths.** This knob sizes the worker thread pool used for sync-activity bodies and for async-activity gRPC response sends. Mixed workloads with long-running sync activities can starve async response delivery (and vice versa) since they share the pool — size to the sum of peak sync activity concurrency and peak in-flight async response sends.
```
sicoyle, line 123: is this a config end users can set?

dapr/ext/workflow/AGENTS.md
```diff
@@ -105,6 +105,26 @@ The entry point for registration and lifecycle:
...
+
+**Concurrency sizing and load characterization.** See `docs/concurrency.md` for sizing recommendations (`maximum_concurrent_activity_work_items`, `maximum_thread_pool_workers`) and an async-vs-sync decision tree. `tests/durabletask/test_async_dispatch_regression.py` (marked `perf`) guards the core invariant: a batch of async activities overlaps on the event loop instead of serializing through the thread pool.
+
+**grpc.aio poller log noise.** The async client can emit benign `BlockingIOError: [Errno 11]` ERROR lines from `grpc.aio`'s `PollerCompletionQueue` under load. It is harmless and retried. `get_grpc_aio_channel` installs an internal `asyncio`-logger filter (`_silence_grpc_aio_poller_noise`) that drops only those records, so the SDK suppresses it automatically with no user action.
```
sicoyle, line 127: do we need this comment?

dapr/ext/workflow/docs/concurrency.md
```diff
@@ -0,0 +1,86 @@
+# Concurrency configuration for `dapr-ext-workflow`
+
+Sizing notes for the worker's concurrency knobs.
+
+## Knobs
```
sicoyle, line 5: are these things end users can set/configure? if so, how?

dapr/ext/workflow/docs/concurrency.md
```diff
@@ -0,0 +1,86 @@
...
+many concurrent async activities, response delivery can serialize through the
+pool and tail latency inflates. Raise `maximum_thread_pool_workers` to widen
+response-delivery throughput.
+
+Mixed workloads with long-running sync activities can starve async response
+delivery (and vice versa) since they share the pool. If that becomes an issue,
+size `maximum_thread_pool_workers` to the sum of peak sync activity concurrency
+and peak in-flight async response sends.
+
+This thread hop goes away when the worker migrates to `grpc.aio`.
```
sicoyle, line 66: what else is left to continue iterating on the async work?

sicoyle (commented):
went through all the files - few comments/questions pls

ext/dapr-ext-workflow/tests/durabletask/test_async_dispatch_regression.py
seherv: The length of the async sleeps used to test that activities are running concurrently and not serially. I'll rename this 😅

tests/ext/workflow/durabletask/test_async_dispatch_regression.py
seherv: A polling loop that waits for the manager to get its first activity in the queue. I'll add a comment

dapr/ext/workflow/AGENTS.md
seherv: Yup, they are parameters for `WorkflowRuntime()`

ext/dapr-ext-workflow/AGENTS.md
seherv: Yeah we do get a very detailed message at `_ActivityExecutor.execute()` explaining that we passed an async function where we shouldn't:

```python
if inspect.iscoroutine(activity_output):
      activity_output.close()
      raise RuntimeError(
          f"Activity '{name}' returned a coroutine on the sync path. "
          f'Declare it with ``async def``, or if it already is, ensure any decorator '
          f'wrapping it uses ``@functools.wraps(fn)`` so the runtime can detect the '
          f'underlying async function.'
      ) 
```

dapr/ext/workflow/AGENTS.md
seherv: Honestly I like using AGENTS.md as a development log of sorts to not waste time rediscovering old issues and rereading the code base. That's just a personal preference though, not sure how much we should let this file grow

dapr/ext/workflow/docs/concurrency.md
seherv: Yeah it's the same ones mentioned in AGENTS.md: `WorkflowRuntime()` takes `maximum_concurrent_activity_work_items`, `maximum_concurrent_orchestration_work_items` and `maximum_thread_pool_workers` as parameters

dapr/ext/workflow/docs/concurrency.md
seherv: The worker still consumes gRPC synchronously in a separate thread. We should use `grpc.aio` instead and replace lots of code that use that thread with simple awaits.

Also the workflow runtime is still sync-first and has its own event loop, and I've had to apply a few patches to make everything run on the same event loop as the user's application. Being able to do `async with WorkflowRuntime()` would avoid having to use those patches.

I'll create issues for both 😄

dapr/ext/workflow/_durabletask/worker.py
seherv: Yup, it was private at least since durabletask got moved into this repo

sicoyle (dismissed):
lgtm - thank you!

acroca (dismissed):

acroca (approved):

marcduiker:
@holopin-bot @seherv sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #1068, opened 2026-05-31
docs(workflow): clarify RetryPolicy docstrings and add max_attempts alias

### LINKED ISSUE #836: [WORKFLOW SDK FEATURE REQUEST] Address many gotchas in workflow activities RetryPolicy
There are multiple gotchas in the workflow activity retry policies as described below:
```
ACTIVITY_RETRY_POLICY = RetryPolicy(
    first_retry_interval=timedelta(seconds=1),
    
   # max_number_of_attempts is the total attempt count, not the retry count
    max_number_of_attempts=5,
    backoff_coefficient=3,
   
    # Set max_retry_interval else there won't be any retries
    max_retry_interval=timedelta(seconds=10),
   
    # DON'T set retry_timeout. If the timeout is hit, the entire workflow fails and the
    # final activity isn't executed.
)
```

## Release Note

RELEASE NOTE: **ADD** Address multiple gotchas in the workflow activity retry policies.

vkbajpaii:
## Description

Addresses #836.

The current `dapr.ext.workflow.RetryPolicy` has several user-facing gotchas (called out in the issue) that are not visible from its docstrings or field names. This PR is the **docs + naming-clarity** slice of that issue; behavior changes are left for follow-up PRs.

## What this PR does

1. **Rewrites the `RetryPolicy` docstrings** in `ext/dapr-ext-workflow/dapr/ext/workflow/retry_policy.py` to spell out:
   - `max_number_of_attempts` counts the **total** number of attempts, not retries — `5` means up to 4 retries after the first attempt.
   - `retry_timeout` failures behave like any other task failure: the surrounding workflow fails unless the call is wrapped in `try` / `except`.
   - `max_retry_interval` only caps the per-attempt delay; retries still happen when it is left as `None`. (The historical "won't retry unless `max_retry_interval` is set" bug was fixed in the now-vendored `_durabletask` engine — verified in `_durabletask/task.py` `RetryableTask.compute_next_delay`, which contains no early bail-out for unset `max_retry_interval`.)
   - Adds a class-level usage example.

2. **Adds a new `max_attempts` keyword-only argument and property** as the preferred name. The legacy `max_number_of_attempts` argument and property remain supported for backward compatibility. Specifying both **or** neither raises `ValueError` to avoid silent inconsistencies.

3. **Adds `ext/dapr-ext-workflow/tests/test_retry_policy.py`** with 11 tests covering construction, validation, the new alias, and the both/neither rejection paths.

4. **Updates the four workflow examples** that construct a `RetryPolicy` (`examples/workflow/simple.py`, `simple_aio_client.py`, `multi-app1.py`, `multi-app2.py`) to use the new `max_attempts` name.

## Acceptance criteria coverage (from #836)

- [x] all field docstrings updated to be clearer on what they do
- [x] `max_attempts` added as the new preferred field; `max_number_of_attempts` kept as a deprecated alias (docs-only deprecation for now — no runtime `DeprecationWarning` to keep this PR backward-compatible and small)
- [x] `retry_timeout` docs updated to indicate the surrounding workflow fails when it fires
- [x] retries apply when `max_retry_interval` is unset — already true in the vendored `_durabletask` engine; the docs now state this explicitly
- [ ] sensible defaults for fields without one — intentionally **out of scope** for this PR (behavior change, backward-compat risk; happy to do as a follow-up PR)

## Testing

```bash
uv run python -m unittest discover -v ./ext/dapr-ext-workflow/tests
# Ran 152 tests in 0.566s — OK

uv run ruff check && uv run ruff format --check
# All checks passed; 332 files already formatted

uv run mypy
# Success: no issues found in 112 source files
```

## Backward compatibility

Existing call sites using `max_number_of_attempts=...` continue to work unchanged. The `max_number_of_attempts` property is preserved and returns the same value as the new `max_attempts` property.

Closes #836 partially (sensible-defaults follow-up tracked in description above).

### DIFF
--- a/examples/workflow/multi-app1.py
+++ b/examples/workflow/multi-app1.py
@@ -25,7 +25,7 @@ def app1_workflow(ctx: wf.DaprWorkflowContext):
 
     try:
         retry_policy = wf.RetryPolicy(
-            max_number_of_attempts=2,
+            max_attempts=2,
             first_retry_interval=timedelta(milliseconds=100),
             max_retry_interval=timedelta(seconds=3),
         )
--- a/examples/workflow/multi-app2.py
+++ b/examples/workflow/multi-app2.py
@@ -29,7 +29,7 @@ def app2_workflow(ctx: wf.DaprWorkflowContext):
     print('app2 - triggering app3 activity', flush=True)
     try:
         retry_policy = wf.RetryPolicy(
-            max_number_of_attempts=2,
+            max_attempts=2,
             first_retry_interval=timedelta(milliseconds=100),
             max_retry_interval=timedelta(seconds=3),
         )
--- a/examples/workflow/simple.py
+++ b/examples/workflow/simple.py
@@ -42,7 +42,7 @@ non_existent_id_error = 'no such instance exists'
 
 retry_policy = RetryPolicy(
     first_retry_interval=timedelta(seconds=1),
-    max_number_of_attempts=3,
+    max_attempts=3,
     backoff_coefficient=2,
     max_retry_interval=timedelta(seconds=10),
     retry_timeout=timedelta(seconds=100),
--- a/examples/workflow/simple_aio_client.py
+++ b/examples/workflow/simple_aio_client.py
@@ -42,7 +42,7 @@ non_existent_id_error = 'no such instance exists'
 
 retry_policy = RetryPolicy(
     first_retry_interval=timedelta(seconds=1),
-    max_number_of_attempts=3,
+    max_attempts=3,
     backoff_coefficient=2,
     max_retry_interval=timedelta(seconds=10),
     retry_timeout=timedelta(seconds=100),
--- a/ext/dapr-ext-workflow/dapr/ext/workflow/retry_policy.py
+++ b/ext/dapr-ext-workflow/dapr/ext/workflow/retry_policy.py
@@ -24,34 +24,95 @@ TOutput = TypeVar('TOutput')
 
 
 class RetryPolicy:
-    """Represents the retry policy for a workflow or activity function."""
+    """Retry policy for workflow activities and child workflows.
+
+    A ``RetryPolicy`` is passed to ``ctx.call_activity()`` or
+    ``ctx.call_child_workflow()`` to automatically retry the task when it
+    fails. The first attempt always runs; on failure, the engine waits
+    ``first_retry_interval`` before the second attempt and multiplies the
+    wait by ``backoff_coefficient`` after each subsequent failure,
+    optionally capped per-attempt by ``max_retry_interval``.
+
+    Example::
+
+        from datetime import timedelta
+        from dapr.ext.workflow import RetryPolicy
+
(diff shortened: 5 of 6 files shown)


kunalworldwide (commented):
Good work on this — the docstrings are significantly clearer now and the `max_attempts` alias is well-handled. A few observations:

**Alias resolution logic**: The mutual exclusion check (raise if both or neither are provided) is clean. One thing to verify: does the durabletask-python library underneath handle `max_number_of_attempts=1` correctly (i.e., no retries, just the initial attempt)? The docstring says "must be >= 1" which implies yes, but worth confirming with a test case for `max_attempts=1`.

**Test coverage**: Good spread — construction with legacy param, new param, both-rejected, neither-rejected, and validation for all fields. The `test_allows_backoff_coefficient_none` case is a nice edge case catch.

**Missing test**: No test for `max_attempts=1` (single attempt, no retries). This is an important boundary case since users might set it expecting "try once, don't retry" behavior.

**Examples updated**: All four example files updated from `max_number_of_attempts` to `max_attempts` — consistent.

**Deprecation notice**: The docstring marks `max_number_of_attempts` as deprecated but there's no `warnings.warn()` call when it's used. Consider adding a `DeprecationWarning` so users see it in their logs and migrate proactively. Not a blocker for this PR though — could be a follow-up.

Overall this is a solid improvement to the developer experience. The docstrings now actually explain the retry behavior instead of just restating the parameter names.

sicoyle:
hey @vkbajpaii thank you so much for your interest and contribution to the Dapr Open Source project and Python SDK! We super appreciate it 🎉 🙌 could you pls resolve the conflicts and copilot feedback and then tag the maintainers/approvers in a comment and we can review? 🙏

CasperGN:
@vkbajpaii can you please look at the conflict?

sicoyle:
> hey @vkbajpaii thank you so much for your interest and contribution to the Dapr Open Source project and Python SDK! We super appreciate it 🎉 🙌 could you pls resolve the conflicts and copilot feedback and then tag the maintainers/approvers in a comment and we can review? 🙏

bump @vkbajpaii 🙏

kunalworldwide (commented):
This is a solid docs-and-naming improvement. The new docstrings make the semantics of `max_attempts` much clearer, and the alias keeps backward compatibility.

A couple of suggestions:

1. `test_rejects_when_both_attempts_fields_supplied` checks that both fields are rejected, but it uses the same value (`3`). Consider adding a test where the two fields have *different* values — that is the case most likely to hide a silent bug if someone accidentally refactors the resolver.
2. The examples were updated from `max_number_of_attempts` to `max_attempts`, which is good. Make sure the public-facing docs site (if this snippet is included) is regenerated as part of the PR, or note it as a follow-up.
3. Please confirm DCO sign-off is present on all commits (`git commit -s`). I did not see the `Signed-off-by` line in the commit list.

The `_resolve_max_attempts` helper correctly rejects both/neither and the property aliases are clean. LGTM once the test edge case is added.

PR #1071, opened 2026-06-01
style: update qs times and wf names

sicoyle:
# Description

I used this to demo at KCD. If you rerun then you get a ton of conflicts on wf names and such. These were some of the tweaks I needed to rerun this properly.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


seherv (commented):
The `>=` is a nice change, I'd do that in other examples too.

nit: I'd use 1 + 10 and explain they're the inputs of the two activities

CasperGN (dismissed):

examples/workflow/simple.py
```diff
@@ -129,10 +129,10 @@ def main():
     wf_client.wait_for_workflow_start(instance_id)
 
     # Sleep to let the workflow run initial activities
-    sleep(12)
+    sleep(20)
 
-    assert counter == 11
-    assert retry_count == 2
+    assert counter >= 11
+    assert retry_count >= 2
```
acroca, line 135: Can you explain this change?
I think it's better to assert a specific value here. By the time the workflow has finished, the counter has to be 11 and the retry count 2, and if they are not it means something undesired happened. It could be hiding something wrong and we should address it. 
If it's flaky we should address the reason of the flakyness and not hide it.

seherv: I was actually thinking `>=` is another way of letting the users know workflow activity execution is at-least-once and not exactly-once. At first the `==` made me think you could get away with activities not being idempotent.

I'm not entirely sure how to address the flakiness issue here. I agree it might be confusing to ship flakey examples but at the same time it also illustrates that workflows are meant to handle flakey code. Maybe an explicit retry policy could help?

acroca: Do you know the reasons of those flaky errors? I think tests like this should be pretty consistent across runs, and I think it's safer to keep it fixed. For example, imagine we introduce a bug that triggers unnecessary retries and the `retry_count` is now always 3 by mistake. With the `>=` we wouldn't get a red CI (maybe another test would fail, but you get the point).

acroca: @sicoyle thoughs on this one?

PR #1081, opened 2026-06-05
GRPC Actors

### LINKED ISSUE #1078: Implement GRPC Actors
Dapr now lets an application host actors entirely **over gRPC**. Instead of the app exposing HTTP callback endpoints that the sidecar calls into, the app opens a single **app-initiated bidirectional gRPC stream** to daprd, registers the actor types it hosts, and receives `invoke` / `reminder` / `timer` / `deactivate` callbacks on that stream. This issue tracks adding gRPC actor host support to this SDK.

Implemented in dapr/dapr: **dapr/dapr#9812** — new alpha RPC `SubscribeActorEventsAlpha1`.

## Why this matters

- **No inbound server port.** An actor host no longer needs to run an HTTP/gRPC server for callbacks — it dials *out* to daprd. Simpler deployment, works in restricted/egress-only networking, fewer moving parts.
- **Consistency across Dapr.** It's the same app-initiated streaming pattern Dapr already uses for streaming PubSub (`SubscribeTopicEventsAlpha1`), configuration watch, and scheduler jobs.
- **One transport for everything.** Invoke, reminders, timers, and deactivation all flow over the single stream, with request/response correlation by id.

## How it works (short version)

1. The app dials **daprd's gRPC port** (the same connection used for the Dapr client).
2. The app opens `SubscribeActorEventsAlpha1` and sends an initial message advertising the actor types it hosts + their runtime config.
3. daprd acks, then **pushes callback requests** onto the stream — each carrying a unique `id`.
4. The app runs the actor code and **replies on the same stream, echoing the `id`** (or sends a typed failure).

No `/healthz` and no `/dapr/config` endpoints are needed: the open stream is the liveness signal, and the initial message replaces the HTTP config endpoint.

## References

- **PR:** dapr/dapr#9812
- **Service / RPC:** `dapr/proto/runtime/v1/dapr.proto` → `SubscribeActorEventsAlpha1`
- **Messages:** `dapr/proto/runtime/v1/actors.proto` → the `SubscribeActorEvents*Alpha1` messages
- **Reference implementation to mirror:** this SDK's existing streaming PubSub subscription (`SubscribeTopicEventsAlpha1`), if present — same stream/oneof/correlation/reconnect shape.

## Implementation brief

A detailed, self-contained brief (protocol contract, lifecycle, requirements, acceptance criteria) is in the collapsible below. It's written to be handed directly to a coding agent (e.g. Claude Code) or used as a spec by a contributor.

<details>
<summary><b>Detailed implementation brief / agent prompt</b></summary>

````text
# Add support for gRPC actors (app-initiated actor callback stream)

## Background

Dapr recently added the ability for an application to **host actors over gRPC**. Previously, an app that hosts actors had to expose HTTP callback endpoints that the Dapr sidecar (daprd) calls into (`PUT /actors/{type}/{id}/method/{method}`, etc.). The new model delivers all actor callbacks over a **single app-initiated bidirectional gRPC stream**, so the app no longer needs to expose any inbound HTTP or gRPC server port for actor callbacks.

This was implemented in dapr/dapr PR #9812 ("Actors: app-initiated gRPC stream for actor callbacks"): https://github.com/dapr/dapr/pull/9812

The goal of this issue is to add **gRPC actor host support** to this SDK.

> How you expose this to end users is entirely your choice. If this SDK already has an HTTP actor model, it would be preferable to keep the user experience consistent with it where it makes sense, since users already understand that model — but the public API shape, naming, and registration ergonomics are up to you. If this SDK does not currently support actors at all, design whatever gRPC-native actor API is most idiomatic here. Do not feel bound to copy any existing HTTP API surface verbatim if a more idiomatic gRPC-native API fits better.

## The key architectural shift (read this first)

With gRPC actors, **the app is the gRPC client**, not a server.

1. The app dials **daprd's gRPC port** — the *same* connection this SDK already uses for the Dapr client (typically from `DAPR_GRPC_ENDPOINT` / `DAPR_GRPC_PORT`, default `127.0.0.1:50001`).
2. It opens the `SubscribeActorEventsAlpha1` stream and sends an initial registration message advertising the actor types it hosts plus runtime config.
3. daprd acks the registration, then **pushes callback requests** (actor method invoke, reminder fire, timer fire, deactivate) onto the stream. Each request carries a unique `id`.
4. The app runs the user's actor code and **sends a response back on the same stream, echoing the `id`** so daprd can correlate it. On failure it sends a typed failure message instead.

No `/healthz`, no `/dapr/config`, no inbound listener: the open stream itself is the liveness signal, and the initial message replaces the HTTP config endpoint.

> **This is the same app-initiated streaming pattern this SDK very likely already implements for streaming PubSub subscriptions (`SubscribeTopicEventsAlpha1`)**, and that Dapr uses for config watch and scheduler jobs. If this SDK supports streaming PubSub, find that existing code and mirror its structure (stream setup, the request/response `oneof`, id correlation, the dispatch loop, reconnect handling). That is your best reference implementation — reuse its conventions.

## The protocol (authoritative contract)

The RPC lives on the main Dapr service in `dapr/proto/runtime/v1/dapr.proto`:

```proto
// App is the client. App-initiated bidirectional stream.
rpc SubscribeActorEventsAlpha1(stream SubscribeActorEventsRequestAlpha1) returns (stream SubscribeActorEventsResponseAlpha1) {}
```

Messages are defined in `dapr/proto/runtime/v1/actors.proto`.

**App → daprd** (`SubscribeActorEventsRequestAlpha1`, a `oneof request_type`):
- `initial_request` (`SubscribeActorEventsRequestInitialAlpha1`) — MUST be the first message. Fields:
  - `repeated string entities` — the actor types this app hosts.
  - `optional Duration actor_idle_timeout`
  - `optional Duration drain_ongoing_call_timeout`
  - `optional bool drain_rebalanced_actors`
  - `optional ActorReentrancyConfig reentrancy` (`{ bool enabled; optional int32 max_stack_depth; }`)
  - `repeated ActorEntityConfig entities_config` — per-actor-type overrides of the above defaults.
  (Unset optional fields mean "use Dapr's default". These map 1:1 to what the HTTP actor config endpoint returns today.)
- `invoke_response` (`SubscribeActorEventsRequestInvokeResponseAlpha1`): `{ string id; bytes data; map<string,string> metadata; bool error; }` — `metadata` carries response headers including `content-type`. `error=true` means `data` is an application-defined error payload that daprd passes back to the original caller verbatim.
- `reminder_response` / `timer_response` (both `SubscribeActorEventsRequestReminderResponseAlpha1`): `{ string id; bool cancel; }` — `cancel=true` stops the reminder/timer from firing again.
- `deactivate_response` (`SubscribeActorEventsRequestDeactivateResponseAlpha1`): `{ string id; }` — ack only.
- `request_failed` (`SubscribeActorEventsRequestFailedAlpha1`): `{ string id; uint32 code; string message; }` — send this INSTEAD of a typed response when the app cannot handle a request. `code` is a gRPC status code (`google.golang.org/grpc/codes`). **`NotFound` (code 5)** is mapped by daprd to a permanent, non-retryable failure (e.g. unknown actor method) — make sure your SDK surfaces a "method/actor not found" condition as `NotFound`. Other errors fall back to `Unknown`.

**daprd → app** (`SubscribeActorEventsResponseAlpha1`, a `oneof response_type`):
- `initial_response` (`SubscribeActorEventsResponseInitialAlpha1`) — empty body, acks registration. (Registration failures arrive as a gRPC error on the stream.)
- `invoke_request` (`SubscribeActorEventsResponseInvokeRequestAlpha1`): `{ string id; string actor_type; string actor_id; string method; bytes data; map<string,string> metadata; }` — `metadata` carries the original caller's request headers, including `content-type` and `Dapr-Reentrancy-Id` when reentrancy is enabled.
- `reminder_request` (`SubscribeActorEventsResponseReminderRequestAlpha1`): `{ string id; string actor_type; string actor_id; string name; string due_time; string period; google.protobuf.Any data; }`
- `timer_request` (`SubscribeActorEventsResponseTimerRequestAlpha1`): `{ string id; string actor_type; string actor_id; string name; string due_time; string period; string callback; google.protobuf.Any data; }`
- `deactivate_request` (`SubscribeActorEventsResponseDeactivateRequestAlpha1`): `{ string id; string actor_type; string actor_id; }`

### Mapping from the HTTP actor callbacks (for intuition, if this SDK has HTTP actors)

| HTTP actor callback                                     | gRPC stream message                          |
|---------------------------------------------------------|----------------------------------------------|
| `GET /dapr/config` (actor registration + config)        | `initial_request` (sent once, app → daprd)   |
| `PUT /actors/{type}/{id}/method/{method}` (invoke)      | `invoke_request` → `invoke_response`         |
| `PUT /actors/{type}/{id}/method/remind/{name}`          | `reminder_request` → `reminder_response`     |
| `PUT /actors/{type}/{id}/method/timer/{name}`           | `timer_request` → `timer_response`           |
| `DELETE /actors/{type}/{id}` (deactivate)               | `deactivate_request` → `deactivate_response` |
| `GET /healthz`                                          | (none — an open stream is the liveness signal) |

The **outbound** actor operations an actor performs (get/set state, register or unregister reminders/timers, invoke another actor) are NOT part of this stream. They already exist as unary RPCs on the Dapr service and you should use those for a fully gRPC-native experience: `GetActorState`, `ExecuteActorStateTransaction`, `RegisterActorReminder` / `UnregisterActorReminder`, `RegisterActorTimer` / `UnregisterActorTimer`, `InvokeActor`. (These are the gRPC equivalents of the HTTP actor state/reminder endpoints.)

## Reference client lifecycle (from the dapr/dapr integration test app)

1. Dial daprd's gRPC address; open `SubscribeActorEventsAlpha1`.
2. Send `initial_request` with the hosted `entities` + config.
3. `Recv()` the first message; assert it is `initial_response` (registration ack).
4. Loop on `Recv()`. For each callback, dispatch it (ideally on its own task/goroutine/thread so the read loop keeps draining), run the user handler, then send the correlated response (or `request_failed`) — **echoing `id`**.
5. **Serialize all stream sends.** gRPC forbids concurrent `Send` on one stream; if you dispatch handlers concurrently, guard sends with a lock/mutex/single writer task.
6. **Reconnect on transient stream errors** (and when daprd isn't ready yet): reconnect with a short backoff and re-send the `initial_request`. Stop when the app is shutting down (EOF / context cancelled).

## What to do (research first, then implement)

1. **Locate this SDK's existing actor implementation, if any** (HTTP actors), and its **existing streaming subscription implementation** (`SubscribeTopicEventsAlpha1` for PubSub), if any. These are your primary references.
2. **Regenerate / update the Dapr protos** so this SDK has the `SubscribeActorEventsAlpha1` RPC and the `SubscribeActorEvents*Alpha1` messages from `dapr/proto/runtime/v1/{dapr,actors}.proto`. Use this SDK's existing proto generation pipeline. If the pinned proto version predates PR #9812, bump it to a commit/release that includes the merged change.
3. **Implement the gRPC actor host**: stream setup, registration from the registered actor types + their config (idle timeout, drain settings, reentrancy, per-type overrides), the dispatch loop, serialized sends, id correlation, error→status mapping (especially `NotFound`), reentrancy header propagation (`Dapr-Reentrancy-Id`), content-type handling, and reconnect.
4. **Wire it into the public API** in whatever way is idiomatic for this SDK. If actor abstractions already exist, reuse them so the same actor classes/handlers can run over either HTTP or gRPC, ideally selected by config.
5. Mark the feature as **alpha/experimental** (the RPC is `Alpha1`).

## Acceptance criteria

- An app can register actor types and host them over gRPC with **no inbound server port** for actor callbacks.
- Actor **method invocation**, **reminders**, **timers**, and **deactivation** all work end-to-end over the stream, with request/response correlation by `id`.
- Handler errors are reported via `request_failed` with a correct gRPC status code; "not found" maps to `NotFound`.
- `content-type` and reentrancy metadata are propagated correctly.
- The stream reconnects automatically after a transient drop / sidecar restart and re-registers.
- Tests: unit tests for message handling/correlation/error mapping, plus an end-to-end test against a real daprd that exercises invoke + reminder + timer + deactivate over gRPC. Follow this SDK's existing actor (or streaming) test patterns.
- Docs/examples updated to show how to run an actor host over gRPC. If this SDK documents HTTP actors, keep the gRPC docs consistent with that style.

## Scope / non-goals

- If this SDK already supports HTTP actors, do **not** remove or change that support; gRPC is additive.
- Do **not** redesign the actor user model if one already exists — reuse existing actor abstractions.
- Keep the public API choice yours; if an HTTP actor experience exists, prioritize consistency with it and with this SDK's idioms.
- If anything in the protocol is ambiguous, the source of truth is `dapr/proto/runtime/v1/actors.proto` and PR #9812 in dapr/dapr.
````

</details>

### PR DESCRIPTION
**Pull request by acroca**
# Description

Adds **alpha** support for hosting Dapr actors over the app-initiated `SubscribeActorEventsAlpha1` gRPC stream (dapr/dapr#9812). The app dials daprd's gRPC port, registers its actor types and runtime config in the stream's first message, and receives all actor callbacks, method invocations, reminders, timers, deactivations, pushed over that single stream. No inbound HTTP or gRPC port is needed for actor callbacks; outbound actor operations (state, reminders, timers, invocations) also go over daprd's gRPC API, so actor apps are fully gRPC-native.

The same actor classes, `ActorRuntime`, and `ActorRuntimeConfig` work unchanged on both transports, only the hosting service differs. HTTP actor hosting (FastAPI/Flask extensions) is untouched.

## What's included

**New public API (alpha)**
- `dapr.actor.ActorGrpcHost` (`dapr/actor/runtime/grpc_host.py`), stream lifecycle: registration handshake (with timeout), concurrent callback dispatch with serialized stream writes and request/response correlation by `id`, automatic reconnect + re-registration after transient drops (`UNAVAILABLE`/`UNKNOWN`/`INTERNAL`, gated on `channel_ready()`), graceful shutdown with in-flight task draining. A stale-session guard ensures replies from handlers that outlive a connection are never sent on its successor.
- `DaprActorGrpcClient` (`dapr/clients/grpc/dapr_actor_grpc_client.py`), implements `DaprActorClientBase` over the unary actor RPCs (`InvokeActor`, `GetActorState`, `ExecuteActorStateTransaction`, reminder/timer registration incl. TTL and failure policies), with `Dapr-Reentrancy-Id` propagation. Closes the long-standing `# TODO: We will allow to use gRPC client later` in `runtime.py`.

**Protocol semantics**
- Error mapping: unknown actor type/method → `request_failed` with `NOT_FOUND` (permanent, non-retryable in daprd); application handler exceptions → `invoke_response{error=true}` carrying the same JSON payload the HTTP extensions return, so callers observe identical errors on both transports.
- Reminder/timer payloads arrive as `BytesValue`-packed `Any`; fire bodies are synthesized to match daprd's HTTP delivery byte-for-byte so `ActorRuntime` needs no changes. Timers registered via the unary gRPC RPC are base64-wrapped by daprd (`json.Marshal` of raw bytes); the host transparently recovers the original payload (documented edge case in `_grpc_callbacks.py`).
- `Dapr-Reentrancy-Id` and `content-type` metadata are propagated in both directions.

**Supporting changes**
- `ActorRuntime.register_actor` gains an additive `actor_client` parameter (defaults to the existing HTTP client).
- `ActorRuntimeConfig`/`ActorTypeConfig`/`ActorReentrancyConfig` gain read-only properties; the registration message is built from typed config objects rather than `as_dict()`.
- `create_aio_channel()` / `resolve_grpc_endpoint()` (`dapr/clients/grpc/_channel.py`), single home for grpc.aio channel setup, now also used by `DaprGrpcClientAsync` (its inline channel construction is replaced).
- `DAPR_REENTRANCY_ID_HEADER` consolidated into `dapr/clients/base.py`.

**Example**, `examples/demo_actor/demo_actor/demo_actor_grpc_service.py` hosts the existing `DemoActor` over the stream; README section added. Note: daprd currently requires a gRPC app channel (`--app-protocol grpc --app-port` with a listener), so the example runs an empty `grpc.aio.server()` on the app port, no actor traffic flows through it.

## Issue reference

Close: https://github.com/dapr/python-sdk/issues/1078

### CODE REVIEW

File: dapr/actor/runtime/grpc_host.py [hunk 1]
```diff
@@ -0,0 +1,508 @@
...
+            response = api_v1.SubscribeActorEventsRequestInvokeResponseAlpha1(
+                id=request.id,
+                data=result,
+                metadata={CONTENT_TYPE_HEADER: JSON_CONTENT_TYPE},
+            )
+        except AttributeError as exception:
+            # Raised by the method dispatcher for unknown actor methods; daprd
+            # maps NOT_FOUND to a permanent, non-retryable failure.
+            await self._send_request_failed(session, request.id, exception)
+            return
```
seherv: If I didn't miss anything in the exception handling chain, this would also swallow an `AttributeError` thrown by the actor itself (not the method dispatch failure), and `_send_request_failed()` is probably not the way you want to handle that.

One solution could be creating an `ActorMethodNotFoundError` exception, raising it from `ActorMethodDispatcher._check_name_exist()` and catching that one here

dapr/actor/runtime/grpc_host.py
```diff
@@ -0,0 +1,508 @@
...
+            )
+            if self._run_task in done:
+                registered_task.cancel()
+                # Surfaces the connection error; a clean exit without
+                # registration can only mean stop() raced start().
+                self._run_task.result()
+        except BaseException:
+            await self.close()
+            raise
+
```
seherv, line 271: We should also handle the case where the task dies *after* a successful registration. Set `self._run_task = None`, log the failure and if it makes sense (probably not?) re-raise the exception

dapr/actor/runtime/grpc_host.py
```diff
@@ -0,0 +1,508 @@
...
+        try:
+            registered_task = asyncio.create_task(self._registered.wait())
+            done, _ = await asyncio.wait(
+                {registered_task, self._run_task}, return_when=asyncio.FIRST_COMPLETED
+            )
+            if self._run_task in done:
+                registered_task.cancel()
+                # Surfaces the connection error; a clean exit without
+                # registration can only mean stop() raced start().
+                self._run_task.result()
```
seherv: I'd log the failure here too, if it's not too spammy

seherv (changes requested):
Nice! 

Just a couple of things that could be cleaned up

sicoyle:
why is this labeled with 1.18 backport?

acroca:
> why is this labeled with 1.18 backport?

Because the feature in the runtime is available in 1.18, so we want it in all 1.18 compatible SDKs.

dapr/actor/runtime/method_dispatcher.py
```diff
@@ -20,6 +20,15 @@
 from dapr.actor.runtime.actor import Actor
 
 
+class ActorMethodNotFoundError(AttributeError):
```
sicoyle: could we have something similar to dapr/actor/error where all of your typed errs could land instead? not sure if the errs are used much in other files or not

dapr/actor/runtime/grpc_host.py
```diff
@@ -0,0 +1,536 @@
...
+        """Routes one callback to the actor runtime and replies on the stream."""
+        callback_type = message.WhichOneof('response_type')
+        if callback_type == 'invoke_request':
+            await self._guard_and_handle(session, message.invoke_request, self._on_invoke)
+        elif callback_type == 'reminder_request':
+            await self._guard_and_handle(session, message.reminder_request, self._on_reminder)
+        elif callback_type == 'timer_request':
+            await self._guard_and_handle(session, message.timer_request, self._on_timer)
+        elif callback_type == 'deactivate_request':
+            await self._guard_and_handle(session, message.deactivate_request, self._on_deactivate)
```
sicoyle: consts pls, or if these are types should we have some static guarantees? There is WhichOneOf, a sort of dispatch table, or the match sort of switch case setup, etc. Ideas below:

```
# built in __init__ since these are bound methods
self._dispatch = {
    'invoke_request': self._on_invoke,
    'reminder_request': self._on_reminder,
    'timer_request': self._on_timer,
    'deactivate_request': self._on_deactivate,
}

handler = self._dispatch.get(callback_type)
if handler is None:
    # unknown variant or nothing set — surface it, don't silently drop
    return
await self._guard_and_handle(session, getattr(message, callback_type), handler)
```
or
```
from typing import assert_never  # 3.11+, else typing_extensions

match callback_type:
    case 'invoke_request':
        await self._guard_and_handle(session, message.invoke_request, self._on_invoke)
    case 'reminder_request':
        await self._guard_and_handle(session, message.reminder_request, self._on_reminder)
    case 'timer_request':
        await self._guard_and_handle(session, message.timer_request, self._on_timer)
    case 'deactivate_request':
        await self._guard_and_handle(session, message.deactivate_request, self._on_deactivate)
    case None:
        ...  # no oneof set
    case _ as unreachable:
        assert_never(unreachable)
```
or
```
expected = {f.name for f in Message.DESCRIPTOR.oneofs_by_name['callback'].fields}
assert set(self._dispatch) == expected  # or the match's case set
```

dapr/clients/grpc/dapr_actor_grpc_client.py
```diff
@@ -0,0 +1,255 @@
...
+        self, actor_type: str, actor_id: str, method: str, data: Optional[bytes] = None
+    ) -> bytes:
+        """Invokes a method on an actor through the InvokeActor RPC.
+
+        The reentrancy id, when present in the current context, is propagated
+        in the request metadata map under the same header name the HTTP
+        transport uses.
+        """
+        # import to avoid circular dependency
+        from dapr.actor.runtime.reentrancy_context import reentrancy_ctx
```
sicoyle: this seems off to me - should this reentracy stuff be moved to correct this and properly setup the pkg instead?

dapr/clients/grpc/dapr_actor_grpc_client.py
```diff
@@ -0,0 +1,255 @@
...
+            actor_id=actor_id,
+            method=method,
+            data=data or b'',
+            metadata=metadata,
+        )
+        response = await self._stub.InvokeActor(request, timeout=self._timeout)
+        return response.data
+
+    async def save_state_transactionally(self, actor_type: str, actor_id: str, data: bytes) -> None:
+        """Saves state through the ExecuteActorStateTransaction RPC.
```
sicoyle, line 112: do we have to worry about if the state store is not transactional? Or do actor state stores already have this check in runtime, therefore making this safe?

examples/demo_actor/demo_actor/demo_actor_grpc_service.py
```diff
@@ -0,0 +1,72 @@
...
+            )
+        ]
+    )
+    ActorRuntime.set_actor_config(config)
+
+    # daprd currently requires a gRPC app channel (--app-protocol grpc --app-port)
+    # before it accepts the actor event stream, so listen on the app port with an
+    # empty gRPC server. No actor traffic arrives here: every actor callback is
+    # delivered over the host's outbound stream.
+    app_server = grpc.aio.server()
```
sicoyle: the PR says no inbound http or grpc port is needed for actor callbacks, and then we stand up an inbound grpc port, so for an end user they will see that we tell them no port needed, but then they must copy an empty server onto the app port... can the ActorGrpcHost own the listener so it never reaches user space instead? Hiding that in the SDK leads to a nicer ex and better dx

examples/demo_actor/README.md
```diff
@@ -123,6 +124,52 @@ expected_stdout_lines:
...
 
+## Run in self-hosted mode with gRPC actor hosting (alpha)
+
+Instead of exposing HTTP callback endpoints, the actor host can receive all
+actor callbacks (method invocations, reminders, timers, deactivations) over a
+single app-initiated gRPC stream (`SubscribeActorEventsAlpha1`). The same
+`DemoActor` implementation runs unchanged; only the hosting service differs.
+
+> This feature is **alpha** and requires a Dapr runtime that supports the
+> `SubscribeActorEventsAlpha1` RPC. daprd must be configured with a gRPC app
```
sicoyle: pls put the runtime version minimum here

examples/demo_actor/README.md
```diff
@@ -123,6 +124,52 @@ expected_stdout_lines:
...
+> channel (`--app-protocol grpc` plus `--app-port`); the service listens on
+> the app port with an empty gRPC server, but no actor traffic flows through
+> it — every actor callback arrives over the host's outbound stream, and all
+> outbound actor operations (state, reminders, timers) use daprd's gRPC API.
+
+1. Run the gRPC Demo Actor service in a new terminal window
+
+   ```bash
+   cd demo_actor
+   dapr run --app-id demo-actor --app-port 3000 --app-protocol grpc -- python3 demo_actor_grpc_service.py
```
sicoyle: can you use the uv run ... instead so we can slowly mv the examples to uv

sicoyle: same for the other cmd below pls

tests/integration/apps/actor_grpc_host.py
```diff
@@ -0,0 +1,135 @@
+"""Actor host over the gRPC actor stream for integration tests.
```
sicoyle: license header pls

tests/integration/test_actor_grpc.py
```diff
@@ -0,0 +1,116 @@
+"""End-to-end tests for actor hosting over the gRPC stream (alpha).
```
sicoyle: license header pls

tests/actor_grpc_utils.py
```diff
@@ -0,0 +1,29 @@
+"""Shared helpers for actor-over-gRPC end-to-end tests."""
```
sicoyle: license header pls

sicoyle (changes requested):
few comments from me for a first pass - thank you!!

dapr/clients/grpc/dapr_actor_grpc_client.py
acroca: the runtime enforces it to be transactional: https://github.com/dapr/dapr/blob/7127d80bb53266c1f90e28b9900ccd4c74fe5739/pkg/actors/state/state.go#L264

dapr/actor/runtime/grpc_host.py
acroca: I went with the `match/case` approach, but with a CI check

dapr/clients/grpc/_channel.py
```diff
@@ -0,0 +1,102 @@
...
+        raise DaprInternalError(f'{error}') from error
+
+
+def create_aio_channel(
+    address: Optional[str] = None,
+    *,
+    interceptors: Optional[Sequence[grpc.aio.ClientInterceptor]] = None,
+    max_grpc_message_length: Optional[int] = None,
+    credentials: Optional[grpc.ChannelCredentials] = None,
+) -> grpc.aio.Channel:
```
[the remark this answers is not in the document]
acroca: `grpc.aio.ClientInterceptor` is exported

dapr/clients/grpc/_channel.py
```diff
@@ -0,0 +1,102 @@
...
+                'grpc.max_receive_message_length',
+                settings.DAPR_GRPC_MAX_INBOUND_MESSAGE_SIZE_BYTES,
+            )
+        )
+
+    uri = resolve_grpc_endpoint(address)
+
+    channel_interceptors: List[grpc.aio.ClientInterceptor] = list(interceptors or [])
+    channel_interceptors.append(DaprClientTimeoutInterceptorAsync())
+    if settings.DAPR_API_TOKEN:
```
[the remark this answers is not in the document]
acroca: `grpc.aio.ClientInterceptor` is exported

tests/examples/test_demo_actor_grpc.py
```diff
@@ -0,0 +1,57 @@
+import pytest
```
sicoyle, line 16: pls add license headers to all of your files missing it

sicoyle (changes requested):
few comments and i can rereview tomorrow :) 

2. Per-actor message_serializer silently ignored (Medium). _get_actor_client builds DaprActorGrpcClient once and caches it, binding actor #1's serializer to every type. Registering two actor types with different serializers gives them all the first one's. HTTP path builds a client per actor, so this is a behavioral regression (invisible with the default JSON serializer).
3. State values re-serialized with the message serializer, not the state serializer (Medium). In dapr_actor_grpc_client._to_transactional_operation, the transactional value is encoded through the single (message) serializer. When state_serializer != message_serializer, persisted bytes differ from the HTTP transport — contradicting the "byte-identical" claim. Related to #2's wiring.
4. Malformed reminder/timer payloads become a retryable error → poison-message loop (Medium). _on_reminder/_on_timer catch everything and route to _send_request_failed; status_code_for_exception maps non-ActorNotFoundError to UNKNOWN, which daprd retries. A structurally invalid payload (JSON decode failure during body synthesis) will redeliver forever. Treat body-decode failures as non-retryable.
5. timer_response populated with the Reminder response type (Low–Medium — verify). _on_timer builds SubscribeActorEventsRequestReminderResponseAlpha1 and assigns it to the timer_response oneof. Deactivate has its own type, so the asymmetry is suspicious — if the proto declares a distinct ...TimerResponseAlpha1, every timer callback raises TypeError. Quick check against the generated proto; a timer test would catch it.
6. App-listener leak on start failure (Low). _start_app_channel_listener assigns self._app_server = server after await server.start(). If add_insecure_port/start() raises (port already bound), the server isn't tracked and close() can't stop it. Assign before start(). (This is the residue of the very thing your comment was about — worth flagging alongside the DX note.)
7. Bad handshake isn't retried (Low). _run_loop only catches TimeoutError/AioRpcError; a malformed ack raises RuntimeError and permanently kills the serving task, whereas an UNAVAILABLE drop reconnects. Inconsistent — decide whether a bad ack should be transient.

acroca:
Addressed your comments, please take a look again.
For your comment #5, the `test_timer_round_trip` constructs that oneof and a type mismatch would fail CI.

sicoyle (approved):
lgtm, pls be sure you cut a new patch release to ensure this makes the 1.18 release with the backport 🙏

PR #1083, opened 2026-06-09
fix(grpc): allow multiple wildcard subscriptions when topic validation disabled

### LINKED ISSUE #436: PubSub disable_topic_validation=True breaks multiple subscriptions
Adding disable_topic_validation=True to pusub subscription raises ValueError when adding multiple subscriptions. Looks like the issue is that the topic map key in this case will only use the pubsub identifier (dapr\ext\grpc\_servicier.py):
```
        if not disable_topic_validation:
            topic_key = pubsub_name + DELIMITER + topic
        else:
            topic_key = pubsub_name
        pubsub_topic = topic_key + DELIMITER
        if rule is not None:
            path = getattr(cb, '__name__', rule.match)
            pubsub_topic = pubsub_topic + path
        if pubsub_topic in self._topic_map:
            raise ValueError(
                f'{topic} is already registered with {pubsub_name}')
        self._topic_map[pubsub_topic] = cb
```

Rishabh-git10:
# Description

Fixes a `ValueError` startup crash and subscription drop when registering multiple wildcard paths on the same pub/sub component with `disable_topic_validation=True`.

Previously, enabling this flag collapsed internal tracking keys down to the pub/sub identifier alone, causing map collisions and rendering the sidecar blind to subsequent subscriptions. This change standardizes discrete internal dictionary keys across all registration paths. It introduces a regex-based pattern matching fallback evaluation inside `OnTopicEvent` and `_handle_bulk_topic_event` to route events when direct string lookups fail.

## Issue reference

Closes #436

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


ext/dapr-ext-grpc/dapr/ext/grpc/_servicer.py
```diff
@@ -197,14 +207,21 @@ def ListTopicSubscriptions(self, request, context):
...
+                if p_name == request.pubsub_name and request.path == path:
+                    if self._match_topic(t_pattern, request.topic) or (
+                        d_val and not any(c in t_pattern for c in ['+', '#', '*'])
+                    ):
+                        cb = w_cb
+                        break
+            if cb is None:
                 context.set_code(grpc.StatusCode.UNIMPLEMENTED)  # type: ignore
-                raise NotImplementedError(f'topic {request.topic} is not implemented!')
+                raise NotImplementedError(f'topic {request.topic} topic not implemented!')
```
seherv: Was this a typo 😅

ext/dapr-ext-grpc/dapr/ext/grpc/_servicer.py
```diff
@@ -81,6 +81,17 @@ def __init__(self):
         self._registered_topics: List[appcallback_v1.TopicSubscription] = []
         self._registered_bindings: List[str] = []
 
+        self._wildcard_topics: List[Tuple[str, str, str, bool, TopicSubscribeCallable]] = []
+
+    def _match_topic(self, pattern: str, topic: str) -> bool:
```
seherv: We need to use a different approach, this would be a second filter on top of what the pubsub already filters by matching a wildcard pattern to a concrete topic name.

More on the following comments

ext/dapr-ext-grpc/dapr/ext/grpc/_servicer.py
seherv, line 163: Here you should build an additional route map based on `topic` to know which handler actually takes it in case you get a wildcard. Something like

```python
if disable_topic_validation and rule is None:
    sub.routes.default = topic
    self._route_map[(pubsub_name, topic)] = cb
```

ext/dapr-ext-grpc/dapr/ext/grpc/_servicer.py
```diff
@@ -197,14 +207,21 @@ def ListTopicSubscriptions(self, request, context):
...
+            cb = self._topic_map[pubsub_topic]
+        else:
+            cb = None
+            for p_name, t_pattern, path, d_val, w_cb in self._wildcard_topics:
+                if p_name == request.pubsub_name and request.path == path:
+                    if self._match_topic(t_pattern, request.topic) or (
+                        d_val and not any(c in t_pattern for c in ['+', '#', '*'])
+                    ):
+                        cb = w_cb
+                        break
```
seherv: Instead of doing the regex match you had before, you can look up the cb on the route map

ext/dapr-ext-grpc/dapr/ext/grpc/_servicer.py
```diff
@@ -293,13 +310,20 @@ def _handle_bulk_topic_event(
...
+            if cb is None:
+                return None  # we don't have a handler
```
seherv: Same as above. Logic looks pretty similar and could be moved to a helper method too

seherv (changes requested):
Thanks a lot for contributing to Dapr! Here are my comments

ext/dapr-ext-grpc/dapr/ext/grpc/_servicer.py
Rishabh-git10: My bad. Looked past it.

Rishabh-git10:
Thanks for the review and guidance @seherv. 
I have consolidated the fallback routing logic into `_get_topic_callback` to remove duplication. Removed the redundant `_match_topic` function.

CasperGN:
@Rishabh-git10 can you fix the conflicts pls?

Rishabh-git10:
Hey, @CasperGN I have resolved the conflicts. Let's see if the checks are happy with it.

Rishabh-git10:
Fixed the routing path keys and the callback overwrite bug. Thanks @acroca.

acroca (approved):

acroca:
LGTM, Thanks for the contribution @Rishabh-git10 :)

marcduiker:
@holopin-bot @Rishabh-git10 sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #1091, opened 2026-06-13
ci: run integration tests against N, N-1, N-2 Dapr runtime versions

### LINKED ISSUE #1067: add test coverage for sdk running with different runtime versions
We're fixing to release 1.18 for Python SDK. There will also soon be a 1.18 Dapr runtime version. We want tests that prove that our SDK version can run compatibly with various runtime versions. Tests that should be added include:

1. Python SDK v1.16 with Dapr Runtime versions: 1.14, 1.15, 1.16
2. Python SDK v1.17 with Dapr Runtime versions: 1.15, 1.16, 1.17
3. Python SDK v1.18 with Dapr Runtime versions: 1.16, 1.17, 1.18

All of our existing tests should be ran for these various versions.

https://docs.dapr.io/operations/support/support-release-policy/#sdk-compatibility

ashoksmore:
# Description

Updates `run-tests.yaml` to validate SDK compatibility against Dapr runtime versions N, N-1, and N-2 per the [SDK compatibility policy](https://docs.dapr.io/operations/support/support-release-policy/#sdk-compatibility).

Changes:
- Adds a `prepare` job that reads the SDK `VERSION` and derives three runtime minors
- Resolves the latest patch (or RC if stable is unavailable) for each minor from `dapr/dapr` releases
- Expands the `validate` job matrix to run existing integration and example tests for each runtime × Python version (3.10-3.14)
- Passes explicit `version` to `setup-dapr-runtime` instead of always using the latest runtime

The same workflow file should work on backported release branches without modification.

## Issue reference

Fixes #1067

## Checklist

* [x] Code compiles correctly: CI workflow change only; `build.yaml` lint/unit checks apply
* [x] Created/updated tests: Reuses existing `tests/integration/` and `tests/examples/` suites across runtime matrix (no new test files; coverage is via CI matrix)
* [ ] Extended the documentation: Not required for this CI-only maintenance change


sicoyle (changes requested):
this is amazing - thank you!!! could you pls mv the python script into its own py file that we can then test and run locally more easily?

acroca:
@ashoksmore Great idea! I was thinking this is potentially useful for other dapr repos too, so I think it'd a good idea to make it reusable.
We could create a new action in https://github.com/dapr/.github/tree/main/.github/actions (this is the repo where we have hared actions and workflows for various dapr repos) that just returns the latest 3 dapr minor versions as an output. 
This new action would be then used together with the `setup-dapr-runtime` like you did in this PR, something like this:
```yaml
  jobs:
    versions:
      runs-on: ubuntu-latest
      outputs:
        versions: ${{ steps.v.outputs.versions }}    # ["1.18.2","1.17.5","1.16.8"]
      steps:
        - id: v
          uses: dapr/.github/.github/actions/dapr-versions@main
    test:
      needs: versions
      strategy:
        matrix:
          version: ${{ fromJson(needs.versions.outputs.versions) }}
      steps:
        - uses: dapr/.github/.github/actions/setup-dapr-cli@main
          with: { version: ${{ matrix.version }} }
```

What do you think?

ashoksmore:
@sicoyle Thanks for your review. I've addressed your feedback. Matrix logic is in tools/compute_compat_matrix.py with unit tests in tests/test_compute_compat_matrix.py.
Ready for another look when you have time, thanks!

ashoksmore:
Hi @acroca, thanks for the suggestion on a shared dapr-versions action in dapr/.github. That makes a lot of sense for reuse across repos.

I kept this PR focused on the python-sdk script per @sicoyle’s feedback, but I’d be happy to help on a follow-up PR in dapr/.github if that’s the direction you’d like to go. Let me know what you prefer.

acroca:
Hey @ashoksmore, I really like the initiative here, but I think there's a gap in the core approach we should sort out before it lands.

The matrix runs the existing tests/integration and tests/examples suites unchanged against N, N-1 and N-2. The problem is those suites already include tests for APIs that only exist in recent runtimes, like conversation (the Llama setup step), jobs, mcp, workflow, crypto and distributed-lock. Running them against 1.17 (which is what N-2 resolves to today, with VERSION at 1.19.0.dev) they'll fail just because the feature didn't exist yet, not because anything regressed. Right now there's no mechanism in the repo to skip a test by runtime version, so the matrix would be red from day one.

The thing we need to solve is making a red job mean something, so it tells us "we broke a supported runtime" and not "this feature is newer than that runtime". The way I read the N-2 guarantee is that existing behavior keeps working on older runtimes, not that new features work on them, but it'd be good to agree on that explicitly.

My preference would be minimum runtime markers, something like `@pytest.mark.min_runtime("1.15")` with a conftest.py hook that reads the runtime version (we can set it as an env in the job, or query /v1.0/metadata) and skips anything above the floor. The nice thing is the set of tests that runs on an older runtime falls out automatically, so there's no separate "core suite" to keep up to date, and each test just declares its floor once. We can also make sure new feature tests carry a marker during review.

One more thing worth putting on the table. This matrix tests the newest SDK against older runtimes. The reverse case, an older SDK against the latest runtime (for example someone on SDK 1.17 upgrading their dapr to 1.19), is probably the more common real world scenario, and it doesn't need any markers because the old tests only touch features that already existed. We could cover that by having the release-1.X branches run their suites against the latest runtime. Pinning release-1.17 to runtime 1.17 specifically isn't worth much, since that pair was already green at release. Not necessarily for this PR, but we should decide which directions we actually want to guarantee.

This is also separate from the shared dapr/.github action idea from before, which is more about where the version resolution lives, so both can land. What do you think?

ashoksmore:
@acroca  Thanks for the detailed feedback; this helps a lot in understanding.

You’re right. The matrix runs the full tests/integration/ and tests/examples/ suites unchanged, and the 1.17 jobs are failing on newer APIs, not because we regressed compatibility with older behavior. I had focused on the CI wiring from #1067 and didn’t account for the fact that many tests target features that didn’t exist on N-2 yet.

I agree with your reading of the N/N-1/N-2 guarantee; existing behavior should keep working on supported older runtimes, not that every new feature must work on them. Without runtime-aware skipping, a red matrix job doesn’t mean much, it mostly says “this test needs a newer runtime.”

I like the @pytest.mark.min_runtime("x.y") approach with a conftest hook (env var from the workflow or metadata lookup). That keeps the suite self-describing and avoids maintaining a separate “core” test list.

Question for you: would you prefer the marker + conftest work in this PR before we enable the matrix, or land the matrix wiring first and follow up with markers in a separate PR?

On the reverse-compat case (older SDK on latest runtime via release branches), that makes sense as a separate decision; happy to help there once we align on direction.

On the shared dapr-versions action in dapr/.github; still happy to contribute a follow-up there too.

Thanks again!

acroca:
> Question for you: would you prefer the marker + conftest work in this PR before we enable the matrix, or land the matrix wiring first and follow up with markers in a separate PR?

Without those, this PR will never get green. We need to get a green build in order to approve the change, so it will have to be added in this PR.

But before adding it, I would like to hear from the other @dapr/maintainers-python-sdk (and @dapr/approvers-python-sdk).

sicoyle:
yes, pls add the markers + conftest work into this PR to support it fully! Thanks for catching that @acroca and for being flexible @ashoksmore!!  I am ok with this change applying to Python SDK first and then once we figure out things like the conftest/markers stuff and have it running successfully (IE we don't think of anything else necessary to roll this out to the rest of the SDKs), then we can generalize this work into the dpar/.github repo as a secondary followup task and come back and update here accordingly. And yes, would love for you @ashoksmore to lead that effort! That would have such a valuable impact on ALLLLLL of the Dapr SDKs which would be amazing!

sicoyle:
> One more thing worth putting on the table. This matrix tests the newest SDK against older runtimes. The reverse case, an older SDK against the latest runtime (for example someone on SDK 1.17 upgrading their dapr to 1.19), is probably the more common real world scenario, and it doesn't need any markers because the old tests only touch features that already existed. We could cover that by having the release-1.X branches run their suites against the latest runtime. Pinning release-1.17 to runtime 1.17 specifically isn't worth much, since that pair was already green at release. Not necessarily for this PR, but we should decide which directions we actually want to guarantee.

@acroca could you please scope out an issue for this? Even for this work, I am happy for us to use Python SDK as the "guinea pig" if you will here to test things out and see the things we need to account for before applying to the rest of the SDKs :)

acroca:
Hey @ashoksmore, will you willing to work on this new decorators approach?

ashoksmore:
Yes, happy to take this on! I’ll add the @pytest.mark.min_runtime markers + conftest hooks in this PR and aim to push an update shortly. Will ping when it’s ready for another look.

PR #1118, opened 2026-07-02
Deprecate cloudevents messages

acroca:
# Description

`dapr.ext.grpc` topic handlers have always received a `cloudevents.sdk.event.v1.Event`. This PR deprecates that in favor of the SDK-owned `SubscriptionMessage`, the same type the streaming subscription API (`DaprClient.subscribe()`) already delivers, as the first step toward removing the `cloudevents` dependency from the `grpc` extra entirely.

## Why

- The SDK uses the `cloudevents` library purely as a data container: daprd does all CloudEvents enveloping/parsing and delivers pre-parsed fields over gRPC. No marshalling code is ever used.
- The type leaks into users' handler signatures, so cloudevents' internal layout is effectively our public API. cloudevents 2.0 moved `cloudevents.sdk.*` to `cloudevents.v1.sdk.*`, which breaks both the extension and every user import (see the dependabot bump #1113). Upstream has moved 1.x to security-fixes-only support.
- The borrowed type also doesn't fit Dapr's delivery model: the topic name was stuffed into `Subject`, and gRPC metadata was crammed into extensions as `_metadata_*` keys. `SubscriptionMessage` has first-class `topic()` / `pubsub_name()`, and gains a `metadata()` accessor here.

## How it works

The delivered type is inferred from the handler's event-parameter annotation, no new parameters on `App()` or `subscribe()`:

```python
from dapr.ext.grpc import App, SubscriptionMessage

app = App()

@app.subscribe(pubsub_name='pubsub', topic='orders')
def new_style(event: SubscriptionMessage):   # annotated → SubscriptionMessage, no warning
    data = event.data()                       # parsed by content type

@app.subscribe(pubsub_name='pubsub', topic='orders2')
def old_style(event):                         # unannotated (or v1.Event) → legacy Event
    ...                                       # + DeprecationWarning at registration
```

## Backward compatibility

- Existing handlers (unannotated or annotated with `v1.Event`) receive the exact same `v1.Event` as before, byte-for-byte, plus one `DeprecationWarning` per subscription at registration time (not per event).
- `SubscriptionMessage.__init__` gains an optional `metadata` parameter, additive; streaming API callers are unaffected (`metadata()` is empty there).

## Note for maintainers: deprecation timeline

The legacy cloudevents handler type is removed by **1.22**:


tests/ext/grpc/test_servicer.py
```diff
@@ -393,3 +393,127 @@ def test_non_registered_job_event(self):
 
 if __name__ == '__main__':
     unittest.main()
+
```
sicoyle, line 430: rename the file to drop that extra i pls so
from: test_servicier.py to test_servicer.py

sicoyle (dismissed):
pls make the one file rename change and then feel free to send this - thank you for making this a very scoped down PR 🙌

acroca:
@sicoyle Done the rename and updated the branch with `main`. Can you re-approve please?

sicoyle (approved):

PR #1126, opened 2026-07-03
Updated gprcio version and regenerate protos

acroca:
- Updated grcpio to `1.81.1`
- Regenerate protos using new versions
- Adds unit test to make sure generated protos match with the floor supported version


sicoyle (approved):
thank you!

PR #1134, opened 2026-07-09
Speed up CI

acroca:
- Faster ollama install
- Reduced waiting times in demo_actor
- Polling for sidecar readyness instead of flat wait.
- Polling for kafka topics instead of flat wait.

[Before](https://github.com/dapr/python-sdk/actions/runs/29000343934/job/86059172194?pr=1135):
image

7m 32s

After (this PR):
image

5m 33s


sicoyle (approved):
amazing - thank you!!

PR #1135, opened 2026-07-09
Fix codecov

acroca:
Codecov has been failing silently for a while now, that's why all PRs get the red check, see: https://github.com/dapr/python-sdk/actions/runs/28903767566/job/85746334110
> error - 2026-07-07 22:44:33,769 -- Upload queued for processing failed: {"message":"Token required because branch is protected"}


sicoyle (approved):
ty!!

PR #1145, opened 2026-07-13
Deprecate reuse-instance-id

acroca:
## Description

Deprecates the `reuse_id_policy` argument on `DaprWorkflowClient.schedule_new_workflow` (sync and async).

Workflow instance IDs can always be reused once the existing instance has reached a terminal state (e.g. COMPLETED, FAILED, or TERMINATED), so the policy is unnecessary, and it currently has no effect anyway, since the underlying proto field is reserved and the policy is never sent on the wire.

The argument is kept for backward compatibility and will be removed in a future release.


sicoyle (approved):
lgtm - thank you!

sicoyle:
@acroca to pls make an issue to roll this out or check this across the other sdks 🙏

PR #1150, opened 2026-07-14
Fix when_all hang when some children complete before construction

### LINKED ISSUE #1149: when_all hangs when some children complete before when_all is constructed
## Expected Behavior

`yield when_all(children)` should complete when all child tasks complete, even if some children finished **before** `when_all(...)` was constructed (deferred fan-in).

Common pattern:

```python
children = [ctx.call_child_workflow(...) for ...]
yield ctx.create_timer(...)          # or an activity
yield when_all(children)             # constructed later
```

## Actual Behavior

If **some but not all** children are already complete when `when_all` is constructed, `when_all` never completes. The orchestration hangs until an outer timeout (if any).

## Root Cause

In `WhenAllTask.__init__`:

```python
super().__init__(tasks)       # CompositeTask counts already-complete children
self._completed_tasks = 0     # resets the counter — drops those completions
self._failed_tasks = 0
```

`CompositeTask.__init__` correctly calls `on_child_completed` for pre-completed children. Resetting the counters afterward loses that count. Later completions only increment from 0, so the total never reaches `len(tasks)`.

Edge cases today:
- 0 early completions → works
- all early completions → works (`is_complete` set before reset)
- partial early completions → **hangs**

## Steps to Reproduce

```python
from dapr.ext.workflow._durabletask import task

children = [task.CompletableTask() for _ in range(5)]
for child in children[:3]:
    child.complete(None)

all_task = task.when_all(children)
# all_task.get_completed_tasks() == 0  (bug; should be 3)

children[3].complete(None)
children[4].complete(None)
# all_task.is_complete is False forever (bug; should be True)
```

Observed in production stress testing against `dapr-ext-workflow==1.18.1` with deferred `when_any([when_all(children), timeout])` after a terminate activity — intermittent depending on how many children finish during the activity.

## Release Note

RELEASE NOTE: **FIX** Bug causing `when_all` to hang when some child tasks complete before `when_all` is constructed

passuied:
Fixes https://github.com/dapr/python-sdk/issues/1149

## Description

`WhenAllTask.__init__` was resetting `_completed_tasks` / `_failed_tasks` after `CompositeTask.__init__` had already counted children that were complete at construction time. That drops early completions, so deferred `when_all(children)` hangs forever when **some but not all** children finished before `when_all` was built.

## Changes

- Stop resetting counters in `WhenAllTask.__init__` (keep empty-list fast path)
- Add unit tests for partial pre-completion, all pre-completion, and `when_any([when_all, timeout])`

## Test plan

- [x] `uv run pytest tests/ext/workflow/durabletask/test_task.py`
- [ ] CI green

RELEASE NOTE: **FIX** Bug causing `when_all` to hang when some child tasks complete before `when_all` is constructed

Made with [Cursor](https://cursor.com)

### DIFF
--- a/dapr/ext/workflow/_durabletask/task.py
+++ b/dapr/ext/workflow/_durabletask/task.py
@@ -385,9 +385,12 @@ class WhenAllTask(CompositeTask[list[T]]):
     """A task that completes when all of its child tasks complete."""
 
     def __init__(self, tasks: list[Task[T]]):
+        # Do not reset _completed_tasks / _failed_tasks after super().__init__.
+        # CompositeTask already initializes them and counts any children that are
+        # already complete via on_child_completed(). Resetting the counters here
+        # drops those completions, so deferred when_all(children) hangs forever
+        # when some (but not all) children finished before when_all was constructed.
         super().__init__(tasks)
-        self._completed_tasks = 0
-        self._failed_tasks = 0
         # If there are no child tasks, this composite should complete immediately
         if len(self._tasks) == 0:
             self._result = []  # type: ignore[assignment]
--- a/tests/ext/workflow/durabletask/test_task.py
+++ b/tests/ext/workflow/durabletask/test_task.py
@@ -62,6 +62,63 @@ def test_when_all_happy_path_returns_ordered_results_and_completes_last():
     assert all_task.get_result() == ['one', 'two', 'three']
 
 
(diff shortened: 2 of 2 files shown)


JoshVanL (approved):

sicoyle (approved):
amazing, thank you!!!

PR #1162, opened 2026-07-16
Pin ports in tests

acroca:
## Description

Fixes the "bind: address already in use" flakes that kill `validate` CI jobs at sidecar startup. The fix is one principle applied everywhere: **tests never let the Dapr CLI pick random ports, and never launch a sidecar until its ports are actually bindable.**

### Root cause 1: the CLI random-port picker assigns duplicate ports

When ports are unset, the CLI picks random free ports independently per listener and can hand the **same port to two listeners of one daprd**. Observed across four listener pairs: metrics vs internal gRPC and metrics vs API gRPC (runs [28439709999](https://github.com/dapr/python-sdk/actions/runs/28439709999), [28198520326](https://github.com/dapr/python-sdk/actions/runs/28198520326), [27741369607](https://github.com/dapr/python-sdk/actions/runs/27741369607), [29419165370](https://github.com/dapr/python-sdk/actions/runs/29419165370)), HTTP vs gRPC ([29482935119](https://github.com/dapr/python-sdk/actions/runs/29482935119)), and metrics vs HTTP ([29485069994](https://github.com/dapr/python-sdk/actions/runs/29485069994)). The last variant is why retry-on-output was abandoned entirely: the metrics listener satisfies the CLI's HTTP readiness probe, the sidecar reports "up and running", the app receives Prometheus text as API responses, and the run exits "successfully" with zero failure markers — structurally invisible to output matching. Worth filing against dapr/cli; this PR stops depending on the picker instead.

### Root cause 2: pinned ports inside the OS ephemeral range get stolen

Dapr's conventional 500xx ports sit inside Linux's ephemeral source-port range (32768–60999). Any process's outbound localhost connection can be assigned one as its *source* port, and an established socket blocks daprd's listener bind. Proven twice on this PR ([validate 3.14](https://github.com/dapr/python-sdk/actions/runs/29478420607), [validate 3.10](https://github.com/dapr/python-sdk/actions/runs/29480618279)): daprd retries the internal bind 10×1s and dies; in the second run the CLI's own preflight reported `Port 50002 is not available` over a minute after every managed process was dead. dapr/dapr#10149 doesn't cover this — it only stops daprd's own dials from stealing its ports. Hence all test-pinned ports live below 32768.

### Root cause 3: teardown races

`dapr run` spawns daprd (and the app, in its own process group) which outlive the CLI, so "CLI exited" never meant "ports free". Rather than tracking process lifetimes, the startup gate verifies the actual invariant — the ports are bindable — which covers every straggler regardless of process-group membership.

### DIFF
--- a/examples/AGENTS.md
+++ b/examples/AGENTS.md
@@ -8,6 +8,7 @@ The `examples/` directory serves as the **user-facing documentation**. Each exam
 2. Tests use a `DaprRunner` helper (defined in `tests/examples/conftest.py`) that wraps `dapr run` commands
 3. `DaprRunner.run()` executes a command and captures stdout; `DaprRunner.start()`/`stop()` manage background services
 4. Tests assert that expected output lines appear in the captured output
+5. The runner pins every sidecar listener port the test didn't set itself and blocks until those ports are bindable before launching. Test-pinned ports must stay below 32768 — see the "Port allocation" section in `tests/integration/AGENTS.md` for the full rationale.
 
 Run examples locally (requires a running Dapr runtime via `dapr init`):
 
--- a/examples/grpc_proxying/README.md
+++ b/examples/grpc_proxying/README.md
@@ -31,8 +31,8 @@ sleep: 5
 -->
 
 ```bash
-# 1. Start Receiver (expose gRPC server receiver on port 50051)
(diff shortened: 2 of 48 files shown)


sicoyle (approved):
LGTM as long as the build passes then this is good to go. THanks!!

PR #1166, opened 2026-07-17
when_all waits for all tasks before surfacing the first failure

acroca:
# Description

`when_all` now waits for every task to finish and then surfaces the first error, instead of completing on the first failure while sibling tasks were still running.

This fixes several rough edges: completions arriving after a failure were silently dropped, so `get_completed_tasks()` and `pending_tasks` reported stale counts from the moment a child failed; and an orchestrator that caught the failure (e.g. to run compensation) would do so while fanned-out activities were still executing. Now the counts stay accurate for the composite's whole lifetime, and by the time the error is raised into the orchestrator, all fanned-out work is known to be finished. When multiple children fail, the first failure is the one reported. The failure is staged internally until the last child completes, so the task never reports `is_failed` while it is still incomplete.

Note this is a behavior change: a fan-out that hits a failure now runs its remaining activities to completion before the workflow fails, instead of failing immediately.

### DIFF
--- a/dapr/ext/workflow/_durabletask/task.py
+++ b/dapr/ext/workflow/_durabletask/task.py
@@ -405,11 +405,17 @@ class CompositeTask(Task[T]):
 
 
 class WhenAllTask(CompositeTask[list[T]]):
-    """A task that completes when all of its child tasks complete."""
+    """A task that completes once all of its child tasks complete.
+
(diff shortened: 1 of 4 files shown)


sicoyle (approved):
LGTM - thank you!!

PR #1167, opened 2026-07-17
Normalize timezone-aware create_timer fire_at to naive UTC

acroca:
Timers created with a timezone-aware `fire_at` are now explicitly normalized to naive UTC in `create_timer` / `create_timer_internal` before the `CreateTimerAction` is built. Previously the scheduled instant was correct only implicitly, relying on protobuf's `FromDatetime` to normalize aware datetimes at serialization time; making the conversion explicit in the SDK guarantees the naive-UTC invariant that the rest of the worker assumes (`current_utc_datetime` is always naive UTC) and protects any future logic that compares `fire_at` against the workflow clock. Also adds executor-level regression tests covering scheduling with an aware `fire_at` (same absolute instant as its naive-UTC equivalent) and clean replay against history recorded in naive UTC.

### DIFF
--- a/dapr/ext/workflow/_durabletask/worker.py
+++ b/dapr/ext/workflow/_durabletask/worker.py
@@ -18,7 +18,7 @@ import random
 import threading
 import warnings
 from concurrent.futures import ThreadPoolExecutor
-from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


sicoyle (approved):
lgtm - thank you!

PR #1178, opened 2026-07-27
Type checking part of lint, and group build/validate statuses

acroca:
- Run `uv run mypy` as part of the lint process because it doesn't require run in a matrix.
- Add a completion CI step so we can mark is as required and it'll group all checks into one requirement.


sicoyle (approved):

yaron2 (approved):

acroca:
This wasn't merged and it's blocking other PRs. Maybe we need to disable the checks in the repo settings so this can be merged and then add them back?

acroca:
I had to disable DCO from the required checks, it doesn't seem to trigger the check in the merge queue.

PR #1201, opened 2026-09-09
Workflow: cross-app client operations

JoshVanL:
Adds an optional app_id argument to every client-level workflow operation on DaprWorkflowClient and its async counterpart: schedule_new_workflow, get_workflow_state, wait_for_workflow_start, wait_for_workflow_completion, raise_workflow_event, terminate_workflow, pause_workflow, resume_workflow and purge_workflow. When set, the operation targets a workflow instance owned by another app in the same namespace, and the target app's WorkflowAccessPolicy decides whether it is permitted. When unset or equal to the local app, the behaviour is unchanged.

The vendored durabletask client carries the value as a TaskRouter with targetAppID on each request, built by the new new_task_router helper. An older runtime ignores the field and applies the operation locally.


dapr/ext/workflow/_durabletask/internal/helpers.py
```diff
@@ -200,10 +200,11 @@ def new_schedule_task_action(
     task_execution_id: str = '',
     propagation_scope: Optional[int] = None,
 ) -> pb.WorkflowAction:
+    # Routing is carried by the enclosing WorkflowAction.router; the inner
+    # ScheduleTaskAction router field is reserved in the protos.
     schedule = pb.ScheduleTaskAction(
         name=name,
         input=get_string_value(encoded_input),
-        router=router,
```
sicoyle, line 206: Heads up — I think this one goes a bit beyond what the PR description covers. Dropping the router off the inner ScheduleTaskAction (and CreateChildWorkflowAction below) changes the wire format for cross-app activities and child workflows, which is separate from the client-level app_id feature this PR is about. Given our N-2 policy, I can't approve this unless every runtime in that window already reads WorkflowAction.router for these actions — if any of them still look at the inner field, cross-app calls break silently against a supported daprd, and the executor tests that would've caught it got relaxed in this PR rather than kept. Could you confirm the earliest runtime version that honors the outer router?

dapr/ext/workflow/_durabletask/internal/orchestrator_service_pb2.pyi
```diff
@@ -322,8 +322,15 @@ class CreateInstanceRequest(_message.Message):
     EXECUTIONID_FIELD_NUMBER: _builtins.int
```
sicoyle, line 322: can you pls confirm there are no manual edits to this file. seems commit 6732be7 includes a manual edit...

tests/integration/test_workflow_cross_app.py
```diff
@@ -0,0 +1,170 @@
...
+
+    caller_client.raise_workflow_event(instance_id, EVENT_NAME, data='finished', app_id=HOST_APP_ID)
+    state = caller_client.wait_for_workflow_completion(
+        instance_id, app_id=HOST_APP_ID, timeout_in_seconds=STATUS_TIMEOUT
+    )
+    assert state is not None
+    assert state.runtime_status.name == 'COMPLETED'
+
+    caller_client.purge_workflow(instance_id, app_id=HOST_APP_ID)
+    assert host_client.get_workflow_state(instance_id) is None
```
sicoyle: every other status check in file polls with a deadline. can you do the same here pls to avoid flake potential in future

tests/integration/test_workflow_cross_app.py
```diff
@@ -0,0 +1,170 @@
+# -*- coding: utf-8 -*-
```
sicoyle, line 1: every docstring here promises the target app's WorkflowAccessPolicy decides whether the operation is allowed, but nothing tests a rejected app_id. That's the security-relevant half of the feature — can we get coverage for the denial case pls?

sicoyle (changes requested):
this is great!! thanks 🚀 few comments so far pls

we already have examples/workflow/multi-app1.py and multi-app2.py showing the context-level cross-app calls. A client-level equivalent would go a long way for discoverability — right now there's no example anywhere showing you can pass app_id straight to schedule_new_workflow.

dapr/ext/workflow/_durabletask/internal/helpers.py
JoshVanL: the field is reserved in the pinned protos, so it cannot be set. No runtime in the N-2 window ever read it. Dapr 1.16.0 ships durabletask-go v0.9.0, 1.17.9 ships v0.11.5, 1.18.0 ships v0.12.1, 1.18.4 ships v0.12.5. In all four, applier.go builds the event with Router: action.Router, the outer field. Dapr itself never reads the inner field on any of those versions or master. The two dropped assertions referenced action.scheduleTask.router, which now raises AttributeError.

dapr/ext/workflow/_durabletask/internal/orchestrator_service_pb2.pyi
JoshVanL: commit 6732be7 is a clean regeneration, not a hand edit. Re-running the script against a clean checkout produces a byte-identical directory

tests/integration/test_workflow_cross_app.py
JoshVanL: I removed the WAP tests bc they require mTLS which is difficult to setup in selfhosted

sicoyle (approved):
LGTM - thank you!!

PR #1202, opened 2026-09-09
Expose a replay-safe new_guid on WorkflowContext

### LINKED ISSUE #1188: [WORKFLOW SDK FEATURE REQUEST] Implement "new_guid" on WorkflowContext
The ability to generate a replay-safe GUID is missing from the API surface of WorkflowContext.

This is a fundamental feature of all Workflow SDKs. e.g. https://github.com/dapr/dotnet-sdk/blob/9a16009d98738fd4f11d1cd7ff58e5afc4fb945a/src/Dapr.Workflow/Worker/Internal/WorkflowOrchestrationContext.cs#L519

magic-peach:
# Description

The underlying orchestration context already generates deterministic GUIDs internally (the worker itself uses this for task execution ids), but nothing on the public `WorkflowContext` exposed it to workflow authors, so anyone needing a stable id inside a workflow had to reach for `uuid4` and break replay determinism.

Adds `new_guid` as an abstract method on `WorkflowContext` and implements it on `DaprWorkflowContext` by delegating to the wrapped context, matching the .NET SDK's `NewGuid`.

## Issue reference

Please reference the issue this PR will close: #1188

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


magic-peach:
@sicoyle could you please review this one as well?

dapr/ext/workflow/_durabletask/worker.py
```diff
@@ -1609,6 +1610,9 @@ def current_utc_datetime(self, value: datetime):
     def is_replaying(self) -> bool:
         return self._is_replaying
 
+    def new_guid(self) -> uuid.UUID:
+        return self.uuid4()
```
JeffreyJPZ:
```suggestion
        return self.new_guid()
```
nit: I think we can use this [alias](https://github.com/dapr/python-sdk/blob/74c3c542f0626197151357e42da34b57c3402d6e/dapr/ext/workflow/_durabletask/deterministic.py#L160) instead for consistency with [here](https://github.com/dapr/python-sdk/blob/74c3c542f0626197151357e42da34b57c3402d6e/dapr/ext/workflow/_durabletask/worker.py#L1664) (not sure why its still called uuid4 though)

JeffreyJPZ (commented):
Looking good! I think this could also benefit from some integration tests in something like `tests/integration/test_workflow_replay_safety.py`. You can see the [dotnet example](https://github.com/dapr/dotnet-sdk/blob/deb05064a4eb4530b9f479ed6e0c6e16ebcc2440/test/Dapr.IntegrationTest.Workflow/ReplaySafetyTests.cs) for inspiration and [other workflow integration tests](https://github.com/dapr/python-sdk/blob/74c3c542f0626197151357e42da34b57c3402d6e/tests/integration/test_workflow_stateful_history.py) for setup, - I can maybe help out if needed.

dapr/ext/workflow/_durabletask/worker.py
```diff
@@ -1609,6 +1610,9 @@ def current_utc_datetime(self, value: datetime):
     def is_replaying(self) -> bool:
         return self._is_replaying
 
+    def new_guid(self) -> uuid.UUID:
+        return self.new_guid()
```
sicoyle, line 1614: this will recurse forever and break call_activity as is I believe. Can you delegate to the mixin's deterministic generator pls

tests/ext/workflow/test_dapr_workflow_context.py
```diff
@@ -54,6 +55,9 @@ def set_custom_status(self, custom_status):
     def get_propagated_history(self):
         return self._propagated_history
 
+    def new_guid(self):
```
sicoyle, line 58: this returns a hardcoded uuid, so nothing exercises the real implementation. Pls add something like:
```
def test_new_guid_is_deterministic_and_unique(self):
    ctx1 = worker._RuntimeOrchestrationContext('abc')
    ctx2 = worker._RuntimeOrchestrationContext('abc')
    g1 = [ctx1.new_guid() for _ in range(3)]
    g2 = [ctx2.new_guid() for _ in range(3)]
    assert all(isinstance(g, uuid.UUID) for g in g1)
    assert len(set(g1)) == 3   # unique within an execution
    assert g1 == g2            # stable across replays
```

sicoyle (changes requested):
great addition so far 🙌  pls see the few comments 🙏

dapr/ext/workflow/_durabletask/worker.py
JeffreyJPZ: Yeah, this one's my bad - should be `super().new_guid()`

PR #1204, opened 2026-09-10
docs(workflow): clarify RetryPolicy field docstrings

magic-peach:
# Description

RetryPolicy's docstrings did not explain three behaviors that are easy to get wrong: max_number_of_attempts counts the first attempt, not just retries after it; leaving max_retry_interval unset means the backoff interval keeps growing without a cap rather than disabling retries; and hitting retry_timeout ends retries with whatever the last attempt returned, it does not run one more attempt first. This spells all three out in the docstrings. Verified each claim directly against the attempt-counting and timeout-scheduling logic in dapr/ext/workflow/_durabletask/task.py before writing it.

## Issue reference

No existing issue tracks this; found while reading the retry policy implementation.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

This is a docstring-only change with no behavior change, so no test or separate documentation update is needed beyond the docstrings themselves.


yaron2 (approved):

magic-peach:
@sicoyle could you please review this one as well?

PR #1215, opened 2026-09-15
Close the event loop invoke_method creates for itself

magic-peach:
# Description

Every synchronous invoke_method call went through get_running_loop failing with RuntimeError, then created a brand new event loop with asyncio.new_event_loop, ran it to completion, and left it open. Each call leaked a whole event loop, visible as an unclosed event loop ResourceWarning. Now the loop is closed in a finally block, but only when this call created it, not when it reused an already running one.

## Issue reference

N/A, self-discovered while reviewing the HTTP invocation client, no existing issue filed.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


tests/clients/test_http_service_invocation_client.py
```diff
@@ -65,6 +65,21 @@ def test_basic_invoke(self):
...
+        import warnings
+
+        with warnings.catch_warnings(record=True) as caught:
+            warnings.simplefilter('always')
+            self.client.invoke_method(self.app_id, self.method_name, '')
+            self.client.invoke_method(self.app_id, self.method_name, '')
+
+        unclosed_loop_warnings = [w for w in caught if 'unclosed event loop' in str(w.message)]
+        self.assertEqual([], unclosed_loop_warnings)
+        self.assertTrue(asyncio.get_event_loop_policy().get_event_loop().is_closed())
```
JeffreyJPZ, line 81: The warning messages are side effects, could we maybe assert on the loop directly by creating it and wrapping this in `with patch("asyncio.new_event_loop", return_value=loop):` or similar?

tests/clients/test_http_service_invocation_client.py
```diff
@@ -65,6 +65,21 @@ def test_basic_invoke(self):
         self.assertEqual(b'STRING_BODY', response.data)
         self.assertEqual(self.invoke_url, self.server.request_path())
 
+    def test_invoke_closes_the_event_loop_it_creates(self):
```
JeffreyJPZ, line 68: Could we also have an equivalent test for the case where invoke fails?

tests/clients/test_http_service_invocation_client.py
```diff
@@ -65,6 +65,21 @@ def test_basic_invoke(self):
...
+        self.server.set_response(b'STRING_BODY')
+
+        import asyncio
+        import warnings
+
+        with warnings.catch_warnings(record=True) as caught:
+            warnings.simplefilter('always')
+            self.client.invoke_method(self.app_id, self.method_name, '')
+            self.client.invoke_method(self.app_id, self.method_name, '')
```
JeffreyJPZ, line 77: Should we split these calls up so we know that each invoke closes its own loop? As it stands this test only checks the loop created by the last invoke (`asyncio.get_event_loop_policy().get_event_loop()` resolves to the event loop for the current thread, which is bound via `asyncio.set_event_loop(loop)` if I'm not mistaken).

JeffreyJPZ (changes requested):
Thanks @magic-peach as always, just have a few comments here

PR #1217, opened 2026-09-17
Add workflow management APIs: list, history and rerun

### LINKED ISSUE #1181: [WORKFLOW SDK FEATURE REQUEST] Expose advanced workflow management APIs
Implement the features (list, history, rewind) in dapr/dapr#9729.

Dapr Agents benefits from the list capability as it enables health endpoints + agent metrics supporting scale to zero (at least until event streaming dapr/proposals#77 lands as it scales better past ~100s of agents).

cc @sicoyle @CasperGN if you have any opinions on the above.

javier-aliaga:
## Description

The vendored durabletask protos already carry `ListInstanceIDs`, `GetInstanceHistory` and `RerunWorkflowFromEvent`, but neither client layer exposed them. This adds all three to `DaprWorkflowClient` and its async counterpart:

| Method | Returns |
|---|---|
| `list_workflow_instance_ids(*, page_size, continuation_token)` | `WorkflowInstanceIdPage` — one page plus the next token |
| `iter_workflow_instance_ids(*, page_size=1024)` | lazy iterator, pages internally (`async for` on the async client) |
| `get_workflow_history(instance_id)` | `list[WorkflowHistoryEvent]` |
| `rerun_workflow_from_event(instance_id, event_id, ...)` | the new instance ID |

All additive; no existing signature changes.

Two behaviours worth knowing:

- `rerun_workflow_from_event` takes the replacement input as one argument. Omit it to keep the original input, pass `None` to clear it. Forwarding code can pass the exported `UNSET` sentinel to mean "not supplied". The engine layer underneath takes the wire's `input` + `overwrite_input` pair directly.
- `WorkflowHistoryEvent` carries `event_id`, `timestamp`, `event_type`, `name`, `task_scheduled_id` and `failure_details`, so a rerun point can be chosen by activity name. Unrecognised event types map to `UNKNOWN` rather than raising.

## Issue reference

Closes #1181
Part of dapr/dapr#9729

## Cross-SDK alignment

Checked against java-sdk#1798 (merged 15 Sep), the only other SDK that implements these — Go exposes durabletask-go's client directly. The page shape (`instanceIds` + `continuationToken`) and the `getInstanceHistory` / `rerunWorkflowFromEvent` surfaces match. Two deliberate differences: the history event carries name, correlation ID and failure details rather than the three fields Java exposes, and the rerun input is one argument instead of Java's `setInput` / `setOverwriteInput` pair.

## Testing

- 59 unit tests across `tests/ext/workflow/test_workflow_management.py` and `tests/ext/workflow/durabletask/test_client_management_apis.py`, covering both clients.
- Full unit suite green; `mypy` and `ruff` clean.
- `examples/workflow/workflow_management.py` runs list → history → rerun against a live sidecar, asserted by `tests/examples/test_workflow.py::test_workflow_management`.


sicoyle:
> rerun_workflow_from_event takes the replacement input as a single argument. Omit it to keep the original input, pass None to clear it.

What does the pass None to clear it actually mean? Does that essentially rerun the workflow from event with a new workflow id so it is essentially technically not a rerun? Can you pls help me understand that one? 🙏

dapr/ext/workflow/_durabletask/aio/client.py
```diff
@@ -307,3 +310,34 @@ async def purge_orchestration(self, instance_id: str, recursive: bool = True):
...
+        res: pb.GetInstanceHistoryResponse = await self._get_stub().GetInstanceHistory(req)
+        return list(res.events)
+
+    async def rerun_orchestration_from_event(
+        self,
+        instance_id: str,
+        event_id: int,
+        *,
+        new_instance_id: Optional[str] = None,
+        input: Union[Any, _Unset] = UNSET,
```
sicoyle: is this the same type for input as the other api calls to trigger a wf?

dapr/ext/workflow/aio/dapr_workflow_client.py
```diff
@@ -306,3 +310,122 @@ async def purge_workflow(self, instance_id: str, recursive: bool = True) -> None
             recursive: The optional flag to also purge data from all child workflows.
         """
         return await self.__obj.purge_orchestration(instance_id, recursive)
+
+    async def list_workflow_instances(
+        self, *, page_size: Optional[int] = None, continuation_token: Optional[str] = None
+    ) -> WorkflowInstanceIdPage:
```
sicoyle, line 438: why not just return a list of workflows? why return a page typespecific obj? is this consistent with the other sdks on this front?

dapr/ext/workflow/dapr_workflow_client.py
```diff
@@ -305,6 +309,124 @@ def purge_workflow(self, instance_id: str, recursive: bool = True):
...
+        events = self.__obj.get_instance_history(instance_id)
+        return [WorkflowHistoryEvent._from_proto(event) for event in events]
+
+    def rerun_workflow_from_event(
+        self,
+        instance_id: str,
+        event_id: int,
+        *,
+        new_instance_id: Optional[str] = None,
+        input: Any = client.UNSET,
```
sicoyle: this is a diff type from other input types in this PR... also unset is a "private sentinel" but it is now exposed directly in the api. can you pls revisit this?

dapr/ext/workflow/workflow_management.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Attributes:
+        instance_ids: The instance IDs in this page, which may be empty.
+        continuation_token: Token to pass to the next
+            :meth:`DaprWorkflowClient.list_workflow_instances` call, or None
+            when this is the last page.
+    """
+
+    instance_ids: list[str]
+    continuation_token: Optional[str]
```
sicoyle, line 71: what is a continuation token? pls clarify with naming further if you can here or in doc string what this means

dapr/ext/workflow/workflow_management.py
```diff
@@ -0,0 +1,194 @@
...
+    CHILD_WORKFLOW_INSTANCE_COMPLETED = 'childWorkflowInstanceCompleted'
+    CHILD_WORKFLOW_INSTANCE_FAILED = 'childWorkflowInstanceFailed'
+    DETACHED_WORKFLOW_INSTANCE_CREATED = 'detachedWorkflowInstanceCreated'
+    TIMER_CREATED = 'timerCreated'
+    TIMER_FIRED = 'timerFired'
+    EVENT_SENT = 'eventSent'
+    EVENT_RAISED = 'eventRaised'
+    CONTINUE_AS_NEW = 'continueAsNew'
+    WORKFLOW_STARTED = 'workflowStarted'
+    WORKFLOW_COMPLETED = 'workflowCompleted'
```
sicoyle, line 113: is there not a way we can pull these from runtime instead of duplicating the strings and creating our own types to represent them here? I worry about drift.

examples/workflow/README.md
```diff
@@ -560,6 +560,62 @@ It shows:
...
+*** history: #-1 WORKFLOW_STARTED name=None
+*** history: #-1 EXECUTION_STARTED name=order_workflow
+*** history: #1 TASK_SCHEDULED name=validate_order [rerunnable]
+*** history: #-1 WORKFLOW_STARTED name=None
+*** history: #-1 TASK_COMPLETED name=None closes=#1
+*** history: #2 TASK_SCHEDULED name=charge_order [rerunnable]
+*** history: #-1 WORKFLOW_STARTED name=None
+*** history: #-1 TASK_FAILED name=None closes=#2 error=amount must be positive, got 0
+*** history: #3 EXECUTION_COMPLETED name=None error=workflow-management-example: Activity task #2 failed: amount must be positive, got 0
+*** rerun started from event #2
```
sicoyle: we expose events, but under the hood its tasks with durabletask. can you update the output here to say event two on task two so it is clear and matches the "rerun started from event #2" line pls

examples/workflow/workflow_management.py
```diff
@@ -0,0 +1,115 @@
...
+    for event in client.get_workflow_history(workflow_instance_id):
+        if event.is_rerunnable and event.name == 'charge_order':
+            return event.event_id
+    raise RuntimeError('no rerunnable charge_order event in history')
+
+
+def main():
+    client = wf.DaprWorkflowClient()
+    wfr.start()
+    sleep(5)
```
sicoyle: why do we need the sleep?

examples/workflow/workflow_management.py
```diff
@@ -0,0 +1,115 @@
...
+    state = client.wait_for_workflow_completion(instance_id, timeout_in_seconds=30)
+    print(f'*** first run finished: status={state.runtime_status.name}', flush=True)
+
+    listed = list(client.iter_workflow_instances())
+    print(f'*** list: instance present={instance_id in listed}', flush=True)
+
+    print_history(client, instance_id)
+
+    charge_event_id = find_charge_event_id(client, instance_id)
+    rerun_instance_id = client.rerun_workflow_from_event(instance_id, charge_event_id, input=25)
```
sicoyle, line 96: what happens if someone inputs 1. an invalid instance id and 2. an invalid change event id?

sicoyle (commented):
thank youuuu!!

did you run checks to see if the changes in this PR are along the same lines in terms of behavior and api surface to that of the other sdks supporting this? 

few comments in general from me pls. Also, from claude below to pls address:

```
Worth settling first

1. UNSET is a private sentinel sitting in a public signature.

dapr_workflow_client.py:382 defaults input to client.UNSET, where client is dapr.ext.workflow._durabletask.client. It isn't in dapr.ext.workflow.__all__, so anyone who wants to name the default has to import from a private module. That matters for wrapper code: a service layer or retry helper that forwards input through can't write input=UNSET, it has to build kwargs conditionally.

This PR already promoted FailureDetails to public for exactly this reason. UNSET deserves the same. Export it from dapr.ext.workflow and reference it in the docstring, or drop the sentinel for an explicit overwrite_input: bool = False pair that mirrors the wire. I prefer exporting the sentinel; the single-argument API reads better and the docstring explains it well.

2. list_workflow_instances returns IDs, not instances.

The return type is WorkflowInstanceIdPage, the iterator yields str, and the RPC is ListInstanceIDs. list_workflow_instance_ids / iter_workflow_instance_ids says what the caller gets, and leaves the unqualified name free if the runtime ever grows a filtered list that returns WorkflowState. I checked go-sdk and nothing there implements these yet, so whatever lands here is the precedent the other SDKs will copy, and renaming post-release is breaking.

On the question in the PR description

Letting NOT_FOUND propagate from get_workflow_history is the right call and I wouldn't invert it. get_workflow_state is the outlier in this client, not the rule: pause, resume, terminate and purge all propagate. And for history specifically, None versus [] would be a genuinely ambiguous return where a raised NOT_FOUND is not. The AGENTS.md wording after the second commit is accurate.

Smaller notes
- task_scheduled_id: getattr(payload, 'taskScheduledId', None) yields None only for payload types that don't define the field. For a type that defines it but leaves it unset you get 0, since it's a proto3 scalar with no presence. The runtime always sets it in practice, but the docstring sits right next to the event_id == -1 convention, which makes None read like the documented "absent" value. Either say so in the attribute docs or note that the field can't be distinguished from a real zero. Don't paper over it with or None.
- iter_workflow_instances loops forever if a sidecar keeps handing back the same non-empty continuation token. You've covered the empty-token case in tests. A one-line guard (stop when the token repeats) closes the remaining hole cheaply. Optional, since a server behaving that way is broken anyway.
- is_rerunnable encodes a server-side rule client-side. The docstring and README both hedge it correctly and test_every_event_type_the_proto_defines_is_mapped will fail loudly on a proto bump, which is the right safety net. I'd keep it. Just be aware that when the runtime adds a fourth rerunnable type, the example's find_charge_event_id pattern silently skips it until someone re-syncs the protos.
- The AGENTS.md "Public API" block at line 79 claims all public symbols are exported and then lists none of WorkflowHistoryEvent, WorkflowHistoryEventType, WorkflowInstanceIdPage or FailureDetails. That list was already stale before this PR, but since you're adding four symbols it's a cheap fix.
- Async side has no equivalent of test_fetches_lazily. The sync generator and the async generator have different laziness semantics in principle, so it's worth the eight lines.
```

dapr/ext/workflow/_durabletask/aio/client.py
javier-aliaga: Same payload type, yes. Both go through shared.to_json into a StringValue.

dapr/ext/workflow/aio/dapr_workflow_client.py
javier-aliaga: Consistent with other sdks, RPC is paginated on the wire

dapr/ext/workflow/workflow_management.py
javier-aliaga: added doc string

dapr/ext/workflow/workflow_management.py
javier-aliaga: I would say no, we can do something with the protos but then we will miss typed enums

examples/workflow/workflow_management.py
javier-aliaga: I will remove it, but other examples are using this sleep

examples/workflow/workflow_management.py
javier-aliaga: It will return NOT_FOUND

javier-aliaga:
> > rerun_workflow_from_event takes the replacement input as a single argument. Omit it to keep the original input, pass None to clear it.
> 
> What does the pass None to clear it actually mean? Does that essentially rerun the workflow from event with a new workflow id so it is essentially technically not a rerun? Can you pls help me understand that one? 🙏

None it meant the activity will receive NONE as input

*** echo_act received 0 (type int)           <- original run, fails
*** echo_act received None (type NoneType)   <- rerun with input=None
*** rerun result = "ok None"

For the second question a rerun is always a new workflow.

dapr/ext/workflow/_durabletask/client.py
```diff
@@ -150,6 +150,46 @@ def new_orchestration_state(
...
+    input alone" and "replace it with null" are only distinguishable through the
+    flag. Callers of the public client express that as one argument; see
+    :data:`dapr.ext.workflow.UNSET`.
+
+    Raises:
+        ValueError: If event_id is negative. The field is uint32 on the wire, so
+        protobuf would otherwise reject it with a message naming neither the
+        argument nor the reason.
+    """
+    if event_id < 0:
```
sicoyle: do we have an upper bound here that we should check as well?

dapr/ext/workflow/__init__.py
```diff
@@ -38,11 +44,16 @@
...
+    'WorkflowHistoryEventType',
+    'WorkflowInstanceIdPage',
+    'UNSET',
     'when_all',
     'when_any',
     'alternate_name',
     'RetryPolicy',
     'TaskFailedError',
+    'FailureDetails',
     'PropagationScope',
```
sicoyle, line 57: can you pls confirm that we must expose all of these? which of these can be internal only?

dapr/ext/workflow/dapr_workflow_client.py
```diff
@@ -305,6 +310,130 @@ def purge_workflow(self, instance_id: str, recursive: bool = True):
...
+            continuation_token: The token from a previous page, to start this
+            page where that one ended. Defaults to starting from the first page.
+
+        Returns:
+            A page of instance IDs, and the token for the next page if there is one.
+        """
+        res = self.__obj.list_instance_ids(
+            page_size=page_size, continuation_token=continuation_token
+        )
+        return WorkflowInstanceIdPage._from_proto(res)
```
sicoyle, line 471: if a user tries a state store that does not support listing keys then this will get an RPC error which we need to catch the error on and provide a nice user facing err for this pls

dapr/ext/workflow/dapr_workflow_client.py
```diff
@@ -305,6 +310,130 @@ def purge_workflow(self, instance_id: str, recursive: bool = True):
...
+        """
+        events = self.__obj.get_instance_history(instance_id)
+        return [WorkflowHistoryEvent._from_proto(event) for event in events]
+
+    def rerun_workflow_from_event(
+        self,
+        instance_id: str,
+        event_id: int,
+        *,
+        new_instance_id: Optional[str] = None,
```
sicoyle, line 517: claudio comment to address pls
```
new_instance_id='' deserves the same guard event_id=-1 just got. I confirmed the presence semantics both ways: newInstanceID has explicit presence, so None leaves it unset and the runtime generates a UUID, while '' sets the field to empty and the runtime takes it literally. That's precisely the failure his own docstring warns about. And the runtime guards the source ID (len(SourceInstanceID) == 0 returns INVALID_ARGUMENT) but not the new one, so nothing downstream catches it.

It's the same class of reachable footgun as -1, with the same cheap fix, and now that one of the two is guarded the asymmetry is odd. Same applies to new_child_workflow_instance_id. Either guard all three or none.
```

sicoyle (commented):
final comments for you - thanks @javier-aliaga 🙌

dapr/ext/workflow/__init__.py
javier-aliaga: I kept most of them, just moved to private this one RERUNNABLE_EVENT_TYPES

dapr/ext/workflow/dapr_workflow_client.py
javier-aliaga: fixed, now we return a clear error

dapr/ext/workflow/dapr_workflow_client.py
javier-aliaga: Done

dapr/ext/workflow/_durabletask/client.py
javier-aliaga: yes, now we also check upper bounds

sicoyle (approved):
LGTM - thanks Javi 🚀

PR #1220, opened 2026-09-18
Fix to_bytes and to_str raising a string instead of an exception

magic-peach:
# Description

`to_bytes` and `to_str` did `raise f'invalid data type {type(data)}'`, which raises a plain string instead of an exception. Python turns this into an unrelated `TypeError: exceptions must derive from BaseException` instead of the intended message. Fixed by wrapping the message in `TypeError`.

## Issue reference

Please reference the issue this PR will close: N/A, self-discovered while reviewing this file, no existing issue filed.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


CasperGN (approved):
Approved.

`raise f'invalid data type {type(data)}'` raises a string, so Python replaces it with `TypeError: exceptions must derive from BaseException` and the intended message never reaches the caller. Wrapping it in `TypeError` is correct, and both `to_bytes` and `to_str` are covered by the new tests.

PR #1221, opened 2026-09-18
Fix convert_from_dapr_duration silently accepting an empty string

magic-peach:
# Description

`convert_from_dapr_duration`'s validity check was `not matched or matched.lastindex == 0`, but `re.Match.lastindex` is `None`, never `0`, when no group matched. So an empty or otherwise ungrouped duration string silently parsed to `timedelta(0)` instead of raising. Fixed by checking `matched.lastindex is None`.

## Issue reference

Please reference the issue this PR will close: N/A, self-discovered while reviewing this file, no existing issue filed.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


CasperGN (approved):
Approved.

`re.Match.lastindex` is `None` when no group participated in the match, never `0`, so the `matched.lastindex == 0` arm was unreachable. Every group in `DAPR_DURATION_PARSER` is optional, so an empty string matched trivially and `convert_from_dapr_duration('')` returned `timedelta(0)` instead of raising `ValueError`. Checking `is None` is the right fix and the added test pins the empty-string case.

Nit, non-blocking: `exeception_context` is misspelled in the new test.

PR #1224, opened 2026-09-20
Guard against a None RpcError details in DaprWorkflowClient.get_workflow_state

magic-peach:
# Description

The sync DaprWorkflowClient.get_workflow_state checks `if 'no such instance exists' in error.details():` with no None check. `RpcError.details()` returns None for connection-level failures such as UNAVAILABLE, where the sidecar is unreachable and there is no trailing metadata from the server. That crashes with an opaque `TypeError: argument of type 'NoneType' is not iterable` instead of the intended "Unhandled RPC error" log and re-raise.

The async client (aio/dapr_workflow_client.py) already guards this correctly with `if error.details() and 'no such instance exists' in error.details():`. This brings the sync client in line with it.

## Issue reference

N/A, self-discovered while auditing workflow client error handling, no existing issue filed. The only related issue is #745 (closed), which is a different failure mode (the original "not found" string match).

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


CasperGN (approved):
Approved.

`RpcError.details()` returns `None` for connection-level failures such as UNAVAILABLE, so `'no such instance exists' in error.details()` raised `TypeError: argument of type 'NoneType' is not iterable` and masked the intended log-and-reraise. The async client in `aio/dapr_workflow_client.py` already guards this the same way, so this removes a divergence between the two clients rather than adding new behaviour. Test covers the UNAVAILABLE path with `details=None`.

